Issue #62, opened 2021-03-31
Merge SNL/ORNL code and CASUS code


RandomDefaultUser:
In GitLab by @RandomDefaultUser on Mar 23, 2021, 16:14

Please feel free to edit/add any points as you see fit!

Before merge:

- [x] Decide on what "MALA" means ("Materials Learning Algorithm" or "Materials Learning and Analysis" or something else entirely?)
- [x] Fix CI and make a pre-merge release (see #61)
- [x] Add SNL/ORNL code here in a way that gives fair credit to the original authors
- [x] Agree on a way to document progress (e.g. google spreadsheet with all the folders that need to be documented, separate issues for seperate functionalitites, etc.)

Merge: 

- [ ] Merge folder by folder 
- [x] Fix open issues in repo (especially investigate #45)

After Merge:

- [ ] Make new release

RandomDefaultUser:
Closed by #168 (amongst others)

PR #168, opened 2021-07-06
Sandia-MALA merge 6/6: Rest

RandomDefaultUser:
Everything else that was not already merged. This should not be much code, and it is probably wise to keep this PR draft until the others merge PRs are merged and then do an rebase. Just so that we do not forget anything.


RandomDefaultUser:
Closes #62 .

PR #196, opened 2021-09-01
Use os.path.join in DataHandler

### LINKED ISSUE #195: DataHandler breaks when paths don't end with "/"
Paths should be joined using os.path.join() or tools from pathlib, not string concatenation.

elcorto:
Paths should be joined using os.path.join() or tools from pathlib. Using
string concatenation is error-prone. For instance, DataHandler code
breaks if paths don't end in "/". I'm sure there is more, only fixed
what bit me.

Closes #195

### DIFF
--- a/mala/datahandling/data_converter.py
+++ b/mala/datahandling/data_converter.py
@@ -1,4 +1,6 @@
 """DataConverter class for converting snapshots into numpy arrays."""
+import os
+
 from mala.common.printout import printout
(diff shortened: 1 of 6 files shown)


RandomDefaultUser:
Very good point. Old habits die hard, I should have done this correctly from the beginning. 
I fixed some more of this in the code. I think I now got most of it.

PR #198, opened 2021-09-01
Add ex05_1 to show how to pass a single input through a network

elcorto:


RandomDefaultUser:
I think this example is a great idea, but could we maybe change the name to keep it consistent with the others? The next one (since #163 uses ex10) would be ex11. What do you think?

PR #210, opened 2021-09-10
Change auth mechanism for deploy steps in workflows

DanielKotik:
We change from using `GITHUB_TOKEN` to a ssh-key pair to enable PRs from forks to actually deploy our GitHub pages - but only when reviewed and approved by members with privilges (maintainer, owner) of mala.

We remove `ACCESS_TOKEN` as is it is no longer needed to access the `test-data` repository (it became public).

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -82,7 +82,6 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: mala-project/test-data
-          token: ${{ secrets.ACCESS_TOKEN }}
           path: mala_data
           ref: v1.0.0
           lfs: false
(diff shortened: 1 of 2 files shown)


RandomDefaultUser (approved):

DanielKotik:
Pushing to the GitHub registry still fails when the review is approved with grafik 

This should be fixed by granting access to the package for forked repositories. However, I can't find my fork in the list:

https://github.com/orgs/mala-project/packages/container/mala_conda_cpu/settings (mala-project → Packages → `mala_conda_cpu` → Package settings (lower right) → Add repository)

This is still open, but not crucial as direct pushes inside mala can trigger trigger the workflow with permission to push the Docker image to the package registry.

RandomDefaultUser (approved):

PR #226, opened 2021-11-02
Parallel descriptors

RandomDefaultUser:
This PR implements parallel descriptor calculation, NOT parallel inference. With this, two things can be achieved:

1. Faster preprocessing
2. Accelerated MD (descriptor calculation split across nodes, network inference on one node+one GPU)

Before merging, some points need to be adressed:

- [x] URGENT: I just found out that my implementation breaks down for large grids / small number of processes. Essentially, the comm.gather() function uses a pickle based mechanism, that has a distinct upper limit on amount of data it can process. Given large enough grids, we exceed this limit. It would be good to replace the implementation with the more performant, yet more complicated comm.Gatherv() function, to be ready for large grids. 
- [x] Currently, at least on my machine I get a segmentation fault when using less then 4 nodes
- [x] Once the code works on 1 node, make sure it is actually fully compatible in the non-MPI case to the current API
- [x] Add a test
- [x] Fix the TODO in the calculator class (ASE automatically detects a running MPI and that causes problems with file I/O)
- [x] Optional: Find out why the parallel MD calculation always stops after 2/3 of the targeted time steps; it doesn't crash, it just goes into a deadlock for no apparent reason. And consistently so after 2/3 of the targeted time steps. Optional because while this is far from good, it does not drastically impair the usage of the feature, one just has to request more time steps, and since on HPC infrastructure this is done regularly (the number of time steps per run is, at least for me, often governed by the time limit of the job), I don't see immediate urgency
- [x] Write issues for things that didn't need immediate fixing:
      - The copy() call (see below)


RandomDefaultUser:
As for the "URGENT" Todo concerning the gathering of descriptors, see e.g. here:

https://stackoverflow.com/questions/36011734/mpi4py-scatter-and-gather-with-large-numpy-arrays
https://stackoverflow.com/questions/37999866/how-to-gather-arrays-of-unequal-length-with-mpi4py

RandomDefaultUser:
I could fix the MPI problem by implementing an alternative. I used Sendv/Recv rather then Gatherv because it seemed wiser given the data structure, yet I am not sure if this is the most efficient way to do that. Someone with better understanding of mpi4py would maybe be able to write a more efficient solution.

RandomDefaultUser:
Further, the solution for the segfault when using low number of nodes was purely in the python side of things; the copy() in b7e02c9ce2574f943c93be5a3ee42ac39368d3f4 fixes it. 
The copy() may be of concern later on. From my understanding, the following happens:
Due to the C<->python interface to lammps, we get the SNAP descriptors as a pointer to some memory. If this chunk is big enough (e.g. small number of nodes, large grids), the content is not explicitly written back to the numpy array, but rather the numpy array serves as a pointer. If we now directly give back the array, the underlying memory will be deallocated, because the original pointer is destroyed, leading to a segfault. This problem is circumvented by explicitly copying the content. 
Now what I am  not sure about is how this is done internally, but it seems to me that for a certain amount of time, our SNAP array will exist twice in memory. This can obviously be a strain on memory. For the time being it works, but before this PR is closed I will create an issue for it.

srajama1:
Can we make sure this workaround is only for small size arrays somehow?

RandomDefaultUser:
The workaround is unfortunately only needed for large size arrays.

RandomDefaultUser:
But you're right, it would be nice if we would only use this when needed. I think in practice, it won't be necessary often, since due to speed we'll most likely be using a number of ranks so high that the local array sizes will be rather small. Once I have done my MD tests and finished up the technical details of this branch I will return to this issue. At best I find a more memory efficient workaround, at worst I'll try to quantify it and only use this workaround when absolutely needed.

srajama1:
:) missed that point. Got it, thanks! The large arrays are the ones where the copy will hurt the most. Hope we don't run into this often. Thanks!

RandomDefaultUser:
I think btw this issues is related to: 
https://github.com/numpy/numpy/issues/18294

RandomDefaultUser:
https://github.com/numpy/numpy/issues/15646#issuecomment-591192650

And this brings even more insight. "Reshape might copy, but in your example will not". Prior to the parallel descriptor code I reshaped the code right in the `__calculate_snap` function, so that caught it. 
What I make out of this is that there is no way around the copy() (or implicit copy() through reshaping) if we don't plan to use the descriptors in the same shape and order we get them. So I think this all kind of depends on whether or not we have to order the descriptors for total energy calculation (for the band energy calculation it doesn't matter, we integrate over the grid anyways). The workflow in this PR definitely does have to reorder because I gather the SNAP descriptors on rank 0 before network passing, but I think the main goal is to eventually also use the TEM in parallel, and then we might not anymore. If that's the case, then we don't have to concern ourselves with the copy() anymore.

RandomDefaultUser:
The test currently does not work in parallel, we eventually should do that. But I think this is an issue for when we do more parallel stuff.

RandomDefaultUser:
I moved the optional task t an issue, creating #233 and #234.

PR #229, opened 2021-11-11
Replace `data_repo_path` modules with env var

elcorto:
Add mala/datahandling/data_repo.py which exposes the same variable as
before, so

-from data_repo_path import data_repo_path
+from mala.datahandling.data_repo import data_repo_path


elcorto:
I added a compatibility layer in `mala.datahandling.data_repo`, not sure if that's the best place. @RandomDefaultUser what do you think?

RandomDefaultUser (approved):
This looks good to me, thanks! :)

PR #230, opened 2021-11-11
Drop python < 3.9 requirement

elcorto:


RandomDefaultUser (commented):
Am I mistaken here or are the changes of #229 appearing in this PR as well? The drop requirement should only affect the requirements.txt, right? The changes in this file look good though.

elcorto:
Yes, that's because that branch here is on top of https://github.com/mala-project/mala/tree/feature-data-repo-path-from-env, pobably one should not send multiple PRs for branches that have a linear history :)

PR #242, opened 2022-01-24
Works locally on test data, now tests on cluster have to be done

### LINKED ISSUE #237: Make DataConverter parallel
The DataConverter should be able to calculate the SNAP descriptors in parallel.

RandomDefaultUser:
Finally closes #237 .
Uses File I/O for LDOS gathering. This is not necessarily the most performant, but the bottleneck are the SNAP descriptors anyway.

### DIFF
--- a/mala/datahandling/data_converter.py
+++ b/mala/datahandling/data_converter.py
@@ -91,7 +91,7 @@ class DataConverter:
         self.__snapshot_description.append(["qe.out", ".cube"])
         self.__snapshot_units.append([input_units, output_units])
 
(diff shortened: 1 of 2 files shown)


RandomDefaultUser:
This works now. Best possible performance for the LDOS would be single node with lots and lots of RAM (since this is just reading from one file to another), and there might be a more performant parallel LDOS parsing algorithm out there. But for now this runs in parallel and fast, I am e.g. able to get both SNAP and LDOS for the aluminium cell with 200x200x200 points in just 20 minutes on 24 nodes.

PR #257, opened 2022-02-12
Fix build docker images workflow

### LINKED ISSUE #254: Fix tagging of Docker images
Currently we have two types of tags for the Docker images, `latest` and `merge`:

grafik

The `merge` image is created on a "PR _review_ that got _approved_ event" and pushed to the registry. However, that is not very useful, since the in the pipeline (for example CPU tests) always the `latest` tagged image is used - which actually isn't the latest image - the `merge` one is!

See action [CPU #264](https://github.com/mala-project/mala/runs/5124360118?check_suite_focus=true) for an example of this misbehaviour.

DanielKotik:
This PR closes #254 and

- fixes/enables caching of Docker layers while building Docker image
- uses `latest` or `x.x.x` tag as Docker tags depending on trigger events (PR, git tag push, push to `master` or `develop`)
- is a first improvement on when to build and how to tag Docker images

Currently, the build process only runs successfully when one of the trigger events is triggered by a member/owner/maintainer from within the MALA repository. A PR from a forked repository is not allowed to push the Docker image built in the workflow to the container registry and fails with the message
```
denied: installation not allowed to Write organization package
Error: Process completed with exit code 1.
```
We need to clarify in a separate issue how the Docker build/push process of PRs should be handled when they come from a fork.

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -1,14 +1,12 @@
 name: CPU
 
 on:
-  pull_request_review:
-    types: [submitted,edited]
   pull_request:
+    # triggered on pull requests to main or develop
     branches:
       - master
       - develop
-
-    # Publish `v1.2.3` tags as releases.
+  push:
     tags:
       - v*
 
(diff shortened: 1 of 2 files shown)


RandomDefaultUser (approved):
Looks good to me, thank you!

PR #266, opened 2022-02-15
Add test to CPU workflow

DanielKotik:
Adds a check if the Conda environment aligns with the dependencies in the `requirements.txt`. If dependencies were added to `requirments.txt`, but were forgotten to be added to the Conda environment, then this test would fail and signal a misalignment between Conda environment in the Docker image and Mala's dependencies (and pssoible extra dependencies).


RandomDefaultUser (approved):
Looks good!

PR #270, opened 2022-02-28
Complete rewrite of the cpu-tests workflow

### LINKED ISSUE #258: Docker build and push to registry guideline
Currently the following process is implemented in the CI:

Every time a push or PR to `master`  or `develop` triggers the [CPU workflow](https://github.com/mala-project/mala/blob/develop/.github/workflows/cpu-tests.yml) pipeline which consists of two job stages, `build-docker-image-cpu` and `cpu-tests`. In the first stage the `latest` tagged Docker image is pulled from the registry and used as a cache reference in the build stage. If layers in the Docker image have changed (e.g. caused by an updated `environment.yml`) the corresponding Docker layers will be rebuilt and the resulting Docker image then pushed to the registry, overwriting the `latest` one. 

We need to think about the following situations and define the desired behaviour:

1. When a PR from _a fork_ is submitted, then pushes to the container registry are forbidden and result in a permission error. That is absolutely correct - we don't want to allow anyone to change the docker images without approval. But we need to find a mechanism for someone external (i.e. via a fork) to run the CI successfully with a modified Docker image, which can then also be updated in the registry when the PR is approved.
2. Do we want to have two separate Docker images (CPU and GPU specific)? It's possible to combine both Conda environments into one (big) image, which has the benefit of only _one_ `build-docker-image` job and not two, almost identical ones for each target environment (from a container perspective, however, each container should cover only one single use case).

The first point has much higher priority than the second.

DanielKotik:
The workflow for the CPU tests now behaves as follows:

### on pull requests to `develop` or `master`

 `build-docker-image-cpu`:
 - check if cache is available and restore `docker-image.tar.gz` if it exists in the cache (Cache key is `github.run_id`, which changes for every new workflow run _except_ for multiple workflow runs within a PRs, e.g. by pressing the re-run workflow button, or when commits are added to the workflow. This avoids building the local Docker image again and again, which most likely is very time consuming. However, if the `cpu-tests.yml` file is repeatedly changed in one and the same PR, this will also result in a new run ids)
 - pull `mala_conda_cpu:latest` image from MALAs container registry at `ghcr.io/mala`
 - build local Docker image (the `Dockerfile` may have changed)
 - compare local built Docker image with latest from registry, if different, save the local one in cache and use it in the subsequent `cpu-test` job
 
  `cpu-tests`:
  - either pull latest image from registry or load local built image from cache
  - run docker container and run `pytest` inside `mala-cpu` Conda environment of Docker container
  
## on push to `develop` or `master` and on pushes with a git tag:
  - same as above plus additional job `retag-docker-image-cpu`
  
  `retag-docker-image-cpu`:
   - if tests succeed, then upload local Docker image to registry and tag with `latest` as well as with a special tag containing the commit sha and branch
   - on push of a git tag, use this tag as a special Docker tag (e.g. git tag `v1.1.0` will result in `mala_conda_cpu:1.1.0`)

This PR closes #258.

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -65,14 +65,14 @@ jobs:
             CACHE=$IMAGE_REPO/$IMAGE_NAME:latest
           fi
 
-          docker build . --file Dockerfile --tag $IMAGE_NAME:latest --cache-from=$CACHE --build-arg DEVICE=cpu
+          docker build . --file Dockerfile --tag $IMAGE_NAME:local --cache-from=$CACHE --build-arg DEVICE=cpu
 
           # Show images
           docker images --filter=reference=$IMAGE_NAME --filter=reference=$IMAGE_REPO/$IMAGE_NAME
 
           # Get image IDs (hash of the local image JSON configuration) and check if images are equal
           IMAGE_ID_OLD=$(docker images --format "{{.ID}}" --no-trunc --filter=reference=$IMAGE_REPO/$IMAGE_NAME:latest)
-          IMAGE_ID_NEW=$(docker images --format "{{.ID}}" --no-trunc --filter=reference=$IMAGE_NAME:latest)
+          IMAGE_ID_NEW=$(docker images --format "{{.ID}}" --no-trunc --filter=reference=$IMAGE_NAME:local)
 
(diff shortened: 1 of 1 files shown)


RandomDefaultUser (approved):
Looks good to me!

PR #277, opened 2022-03-07
Avoids pipeline trigger on approved reviews:

### LINKED ISSUE #212: Pushing to the GitHub registry still fails when the review is approved
Pushing to the GitHub registry still fails when the review is approved with grafik 

This should be fixed by granting access to the package for forked repositories. However, I can't find my fork in the list:

https://github.com/orgs/mala-project/packages/container/mala_conda_cpu/settings (mala-project → Packages → `mala_conda_cpu` → Package settings (lower right) → Add repository)

This is still open, but not crucial as direct pushes inside mala can trigger trigger the workflow with permission to push the Docker image to the package registry.

See also: https://github.community/t/new-pakckages-access-via-github-token-impractical-for-forks/170495

_Originally posted by @DanielKotik in https://github.com/mala-project/mala/issues/210#issuecomment-916808492_

DanielKotik:
If the review is approved, a run of the pipeline was triggered but
failed due to insufficient rights from the forked repository. There is
actually no need for an extra run of the pipeline once a review is
approved since a push to master or develop is likely to follow. This
merge/push is done with sufficient rights of the repositories owners.

Fixes #212

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,8 +1,6 @@
 name: docs
 
 on:
-  pull_request_review:
-    types: [submitted,edited]
   pull_request:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


RandomDefaultUser (approved):
Looks good to me!

PR #280, opened 2022-03-12
Reactivate sphinx' viewcode extension

DanielKotik:
Since we have long been public, we can reactivate the `viewcode` extension again, which allows to jump to the code from within the docs.


RandomDefaultUser (approved):
You're absolutely right, that's a good idea!

PR #286, opened 2022-03-22
Fix examples

### LINKED ISSUE #275: Failures in several examples
I ran the examples (same software setup as in #273). There are multiple failures. I won't list them individually but just attach a log file
[out_stripped.txt](https://github.com/mala-project/mala/files/8172913/out_stripped.txt).

Seems like those were not run for some time? I suggest to integrate at least some of them into the test suite (e.g. a test that executes them would be sufficient).

One however stood out: `ex04_hyperparameter_optimization.py` ran forever, I killed it after 30 min at output:
```
2022-03-02 21:45:15,107 Trial 4729 pruned.
2022-03-02 21:45:15,860 Trial 4730 pruned.
2022-03-02 21:45:16,695 Trial 4731 pruned.
```

Seems like there is no termination criterion.

RandomDefaultUser:
Closes #275 .

I fixed all the small bugs that @elcorto found. I have further added a test "test_examples.py" in the examples directory. I have deliberately chosen not to include those in the regular test suite / CI, as the functionalities of the examples are reproduced by the test suite - having the CI run those twice would unnecessarily increase the runtime. The bugs in this case always came from me having modified the examples for some development and then not testing properly thereafter. My idea would be that the "test_examples.py" can work more as a check for developers, and I have included a note in the CONTRIBUTE.md.


elcorto:
Looks good!

Not duplicating test coverage is a good point. Having said that, examples are the first contact point for new users (also students) so I think having them tested frequently is important. If devs are aware of that, then running them manually is probably ok, even though history has shown that not enforcing this leads to things getting out of sync. But project is new and quite in flux, so people should expect rough egdes.

On the technical side, we may think about using `pytest.mark` to mark them as "examples" such that we can run them in a pytest-ish way in the future. We could then move `test_examples.py` to `test/` and have the CI run `pytest -m "not examples"` with having them run by default (just call `pytest`), or not including them by default and let devs do `pytest -m "examples"` manually.

One thing: since you use `runpy`, should we add that to the test deps?

RandomDefaultUser:
Thanks for the feedback @elcorto! I really like the idea with the `mark` option. That way we can, in a later PR, modify the CI to only run examples let's say when something is merged onto develop, but not during the PR. And locally, developers will more likely run  `pytest` itself and test both examples and tests locally. I have added that in the last commit!

As to `runpy`, isn't that part of the core python package? So installed by default?

elcorto:
Nice work with `mark` here. Note that you can decorate classes in `pytest`. Did this quickly in 85ff8400ef2bf6b74eb758271bc4b656b8a6d302 .

`runpy`: Of course, that's in the standard lib, my mistake.

RandomDefaultUser:
Ah cool, great, thanks for the hint!

elcorto (approved):
The branch looks good now.

PR #288, opened 2022-03-22
grid spacing based -> voxel based

RandomDefaultUser:
Except for simpson/trapezoid grid integration, everything works for non-cubic cells now.


RandomDefaultUser:
I tested this both on Al and Be, both for predicting DOS and total energy from the LDOS. The new implementation gives the same results up to neV/atoms, for Be I think it was machine precision even. So this works well, and I will merge it.

PR #290, opened 2022-03-23
Fixed `extra requires` (small typo)

### LINKED ISSUE #260: Fix `extra_requires`
In two different local environemnts, and the CI, the command `pip install -e .[opt]` results in `WARNING: mala 0.2.0 does not provide the extra 'opt'`. I am unsure why. The most recent setuptools version this was tested with is 59.5.0 (because the current torch version breaks with versions higher then that). 

Comment by @DanielKotik:
Once this is fixed we need to make the approbriate changes in
```yml
# cpu-tests.yml
- name: Install mala package
  # ...
  run: |
     # ...
     # install extra dependencies
     pip install oapackage  # change to → pip install -e .[opt]
     pip install pytest     # change to → pip install -e .[test]
```
This then would also serve as a test indirectly.

RandomDefaultUser:
Closes #260 .

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -155,11 +155,7 @@ jobs:
           conda env export -n mala-cpu > env_1.yml
 
           # install mala package
-          pip --no-cache-dir install -e .
-
-          # install extra dependencies
-          pip install oapackage
-          pip install pytest
+          pip --no-cache-dir install -e .[opt, test]
 
       - name: Check if Conda environment meets the specified requirements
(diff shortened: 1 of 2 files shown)


DanielKotik:
@RandomDefaultUser  It is not obvious to me what the small typo is... What change exactly resolves the issue?

RandomDefaultUser:
In the setups we had written e**tx**ras_require=extras instead of e**xt**ras_require=extras :sweat_smile:

DanielKotik (approved):
I was unable to spot this small typo. Good catch @RandomDefaultUser !

PR #293, opened 2022-03-24
MALA on Windows

### LINKED ISSUE #56: Support Windows OS?
In GitLab by @RandomDefaultUser on Feb 26, 2021, 19:44

We might have to clarify on what platform our collaborators are developing. I'm not sure if our code works on Windows systems right away, or if there is need to change, e.g. file path issues (its always good practice to use the os independent path module `os.path` from the python standard library anyway, e.g. `os.path.join`).

- [x] check if there is need to support windows
- [x] if so, check if fesl is windows compatible and fix issues if not

RandomDefaultUser:
The installation itself is trivial, as the default MALA installation only needs a few readily available packages, and puts everything else in the hand of the user. 
I had to alter one test, because through writing this I realized that `oapackage` was required for the test suite to run, which should not be the case. 
I also created a `make.bat` so the documentation may be built locally on Windows. 
Tested on a recent Windows 10. 

Closes #56 .

### DIFF
--- a/docs/make.bat
+++ b/docs/make.bat
@@ -1,0 +1,26 @@
+@echo off
+
+SET SOURCEDIR=source
+SET BUILDDIR=_build
+@REM Should be equivalent to the ?= in the Makefile
+IF "%SPHINXBUILD%"=="" (SET SPHINXBUILD=sphinx-build)
+@REM We currently don't have any other SPHINXOPTS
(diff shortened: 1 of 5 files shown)


DanielKotik:
Cool!

DanielKotik (approved):

PR #302, opened 2022-03-28
Remove pandas (EDIT: ADDING pandas)

RandomDefaultUser:
I tested it locally, and conda did not uninstall any other package. It tried to upgrade openssl and the ca-certificates, but since it was an upgrade, that seems unrelated.

### DIFF
--- a/install/mala_cpu_base_environment.yml
+++ b/install/mala_cpu_base_environment.yml
@@ -11,5 +11,4 @@ dependencies:
   - ase
   - pytorch-cpu
   - mpmath
-  - pandas
   - tensorboard
--- a/install/mala_cpu_environment.yml
+++ b/install/mala_cpu_environment.yml
@@ -94,7 +94,6 @@ dependencies:
   - openssl=1.1.1l=h7f98852_0
   - optuna=2.10.0=pyhd8ed1ab_0
   - packaging=21.3=pyhd8ed1ab_0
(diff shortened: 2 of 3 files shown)


DanielKotik:
@RandomDefaultUser Please check what causes the pipeline to fail:

```
=========================== short test summary info ============================
FAILED test/hyperopt_test.py::TestHyperparameterOptimization::test_hyperopt
FAILED test/hyperopt_test.py::TestHyperparameterOptimization::test_distributed_hyperopt
======= 2 failed, 24 passed, 7 skipped, 40 warnings in 195.49s (0:03:15) =======
Error: Process completed with exit code 1.
```

RandomDefaultUser:
Ugh, yeah, it is. It all comes back to me now: We don't need pandas, but optuna does. It is one of the optional dependencies, but since we don't need all of the optional dependencies (things like mxnet, mpi4py, etc.) I chose to include this manually. I will check tomorrow if maybe one of the optuna installation options installs pandas without installing anything unnecessary. That would be optimal in my opinion, since then it would be clear that while we don't need pandas, optuna does. Elsewise I will leave a comment this time, so it is not forgotten again.

RandomDefaultUser:
As it turns out, the best thing we can do atm is to simply add `pandas` fully again, elsewise we would have to install quite a few extra dependencies through optuna.

DanielKotik (approved):

PR #305, opened 2022-03-30
Add workflow for publishing on PyPI and TestPyPI

DanielKotik:
I've added secrets `TEST_PYPI_API_TOKEN` and `PYPI_API_TOKEN` to the mala settings.

On pushes to `master` and `develop` this workflow gets triggered and tries to build the mala package. On tags being pushed, the package will also be uploaded to TestPyPI. Only when a tag is turned into a release via GitHub, the package is published on PyPI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ This repository is structured as follows:
 
 ## Installation
 
+> **WARNING**: Even if you install MALA via PyPI, please consult the full installation instructions afterwards. External modules (like the QuantumESPRESSO bindings) are not distributed via PyPI!
+
 Please refer to [Installation of MALA](docs/source/install/README.md).
 
 ## Running
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,59 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+
+on:
+  push:
+    branches:
+      - master
+      - develop
+    tags:
+      - 'v*.*.*'
+  release:
+    types:
+      - published
+
+jobs:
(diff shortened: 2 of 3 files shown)


DanielKotik:
@RandomDefaultUser Please change `setup(name='mala', ...)` in `setup.py` to a reasonable name that is not already taken on PyPI. Any suggestions?

RandomDefaultUser:
Some suggestions for the name:

1. `mala-core`: it may be that we eventually build more and more software around the `mala` ecosystem, e.g. an additional automation layer for HPC, so this would be the "core" package
2. `mala-dft`: since we mostly work with DFT data; technically that's not exclusive, I am aware of that. 
3. `mala-ml`: since we work with machine learning, and the other `mala` package does not
4. `mala-ms`: since we are concerned with materials science
5. `materials-learning-algorithms`: this may be excessively long

I would advocate for 1.

Zevrap-81:
if you want something that sounds good then maybe "pymala"?

README.md
```diff
@@ -23,6 +23,8 @@ This repository is structured as follows:
 
 ## Installation
 
+> **WARNING**: Even if you install MALA via PyPI, please consult the full installation instructions afterwards. External modules (like the QuantumESPRESSO bindings) are not distributed via PyPI!
```
DanielKotik, line 26: Is this `>` before `**WARNING**` of any value, or just a typo?

DanielKotik (commented):
see comment

README.md
RandomDefaultUser: It's a blockquote, per this suggestion: https://stackoverflow.com/a/55850542

DanielKotik: You're right.

PR #309, opened 2022-04-10
First adjustments for parallel TEM

RandomDefaultUser:
Parallel TEM is not yet working with this, but this should clear things from the python side and solidifies the API.


RandomDefaultUser:
I am closing this PR here. A direct merge to develop seems a bit messy, when #315 and #318 also have to be merged. I will create a new branch to first converge these three branches and then merge to develop.

PR #315, opened 2022-05-02
Added new factory methods and caching algorithms

RandomDefaultUser:
This PR provides several factory methods for observables and adds functionality to cache intermediate results (e.g. the number of electrons, Fermi energy). It enhances speed and usability. As a (very small) downside, the standard way of accessing calculators is thus now very streamlined, i.e., if you use LDOS.number_of_electrons rather then LDOS.get_number_of_electrons(...) values like temperature, voxel dimensions, etc. will be assumed to be internally correct. This is the case in 99.9% of use cases though, so no big deal.

Before merging I have to do a full scale test on some production level data and update the docs. The bare functionality (save bugs, although the CI reports no errors at the moment) is there however. 

Closes #308

PR #318, opened 2022-05-19
Make Angstrom the default internal length unit

### LINKED ISSUE #30: Clean up length units
In GitLab by @RandomDefaultUser on Jan 14, 2021, 10:57

I just realized that out of convenience I left all length units in Bohr. So while energy units are following ASE notations by being in eV, length units do not. We should make the switch to Angstrom soon.

RandomDefaultUser:
In doing so, we keep consistent with ASE and have an existing framework for working with other codes.
Closes #30 .


RandomDefaultUser:
I am closing this PR here. A direct merge to develop seems a bit messy, when #315 and #309 also have to be merged. I will create a new branch to first converge these three branches and then merge to develop.

PR #348, opened 2022-09-26
Re-integrating TEM into MALA main code 

RandomDefaultUser:
Installation has been tested.

### DIFF
--- a/docs/source/install/external_modules.rst
+++ b/docs/source/install/external_modules.rst
@@ -21,11 +21,8 @@ Build Quantum ESPRESSO
   (the total energy module is based on version 6.4.1): ``git checkout qe-6.4.1``
 * Make sure you have an (MPI-aware) F90 compiler such as ``mpif90`` (e.g.
   Debian-ish machine: ``apt install openmpi-bin``, on an HPC cluster something
-  like ``module load openmpi gcc``). In fact the extension module that we build
-  doesn't use MPI at the moment, so a non-MPI compiler should also work (e.g.
-  ``gfortran`` or ``ifort``). In any case, make sure to use the same compiler
-  for QE and the extension (``--f90exec=`` in
-  ``build_total_energy_energy_module.sh``).
+  like ``module load openmpi gcc``). Make sure to use the same compiler
+  for QE and the extension (``--f90exec=`` in ``build_total_energy_energy_module.sh``).
 * We assume that QE's ``configure`` script will find your system libs, e.g. use
   ``-lblas``, ``-llapack`` and ``-lfftw3``. We use those by default in
   ``build_total_energy_energy_module.sh``. If you have, say, the MKL library,
@@ -41,7 +38,7 @@ Build Quantum ESPRESSO
 * Change to the ``q-e`` directory
 
   * If you already have a build of QE, go into the ``q-e`` folder and run ``make veryclean``.
-* Run ``./configure CFLAGS="-fPIC" CPPFLAGS="-fPIC" FFLAGS="-fPIC -fallow-argument-mismatch" foxflags="-fPIC"``
+* Run ``./configure CFLAGS="-fPIC" CPPFLAGS="-fPIC" FFLAGS="-fPIC -fallow-argument-mismatch"``
 * Run ``make all`` (use ``make -j<your number of cores> all`` for a faster
   compilation process).
 * Change back to the  ``external_modules/total_energy_module`` directory of the
--- a/docs/source/install/external_modules.rst
+++ b/docs/source/install/external_modules.rst
@@ -33,7 +33,7 @@ Build Quantum ESPRESSO
   * GNU compiler specific: we use ``-fallow-argument-mismatch``
 * Change to the ``external_modules/total_energy_module`` directory of the
   MALA repository
-* Run the script ``prepare_qe.sh -q /path/to/your/q-e`` with ``/path/to/your/qe``
+* Run the script ``prepare_qe.sh /path/to/your/q-e`` with ``/path/to/your/qe``
   being the path to the ``q-e`` directory
 * Change to the ``q-e`` directory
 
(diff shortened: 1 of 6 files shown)


external_modules/total_energy_module/prepare_qe.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+while getopts q: flag
```
elcorto, line 3: This script will silently fail and do nothing if `-q ...` is *not* provided. Then it will execute `cd` which is `cd ~` and there do nothing. It may be better to make this a required argument.

Or use `set -u` at the start to fail on undefined variables such as `$root_dir`.

external_modules/total_energy_module/prepare_qe.sh
```diff
@@ -0,0 +1,14 @@
...
+do
+    case "${flag}" in
+        q) root_dir=${OPTARG};;
+    esac
+done
+
+cd $root_dir
+if [ "$(grep -q "$foxflags -fPIC" install/configure)" == "" ]; then
+  echo "HERE"
```
elcorto, line 12: Is this a "check" to make sure that the condition is met? If so then maybe make the script fail if grep doesn't find what it is supposed to find.

docs/source/install/external_modules.rst
```diff
@@ -2,21 +2,63 @@ External modules
...
+-----------------------
+
+* Clone QE source code: ``git@gitlab.com:QEF/q-e.git``
+* Chage into the ``q-e`` directory and check out the correct branch
+  (the total energy module is based on version 6.4.1): ``git checkout qe-6.4.1``
+* Make sure you have an (MPI-aware) F90 compiler such as ``mpif90`` (e.g.
+  Debian-ish machine: ``apt install openmpi-bin``, on an HPC cluster something
+  like ``module load openmpi gcc``). In fact the extension module that we build
+  doesn't use MPI at the moment, so a non-MPI compiler should also work (e.g.
+  ``gfortran`` or ``ifort``). In any case, make sure to use the same compiler
```
elcorto, line 26: These are the instructions we wrote back in the day. I'm not up to date re. the build process but just to make sure: is it still true that the extension module is serial?

external_modules/total_energy_module/prepare_qe.sh
RandomDefaultUser: This only checks if the foxflags command has already been added to the configure file. I think users may accidentally execute the script multiple times, so I added a check. If the lines are already there all is well and good :)

external_modules/total_energy_module/prepare_qe.sh
RandomDefaultUser: That's a very good point I totally overlooked. I fixed this in my last commit.

docs/source/install/external_modules.rst
RandomDefaultUser: You're right, this is outdated. We now build in parallel. I have fixed this line.

external_modules/total_energy_module/prepare_qe.sh
elcorto: That's not exactly what I meant. I didn't even see that a colon was not used :)

Not using `q:` (without `:`) meant the option did not require an arg, so your test runs of the script should all have failed since the path given to `-q <path>` was never used, strange.

Anyway what I meant was to make this path a required argument (`script.sh foo`) rather than an option (`script.sh -q foo`), i.e.

```sh
#!/bin/sh

set -eu

err(){
    echo "error: $@"
    exit 1
}

[ $# -eq 1 ] || err "need exactly one argument" && root_dir=$1

echo ">>$root_dir<<"
```

If you want to go the option route, then make sure to place any `set ...` *at the top*. In particular `set -u` will make the script fail when `-q <path>` is *not* given since then `$root_dir` is not defined.

```sh
#!/bin/sh

set -eu

err(){
    echo "error: $@"
    exit 1
}

while getopts q: opt; do
    case $opt in
        q) root_dir="${OPTARG}";;
    esac
done

# Since we use set -u (error out on undefined vars, we actually die before this
# check.
[ -z "$root_dir" ] && err "root_dir not defined"

echo ">>$root_dir<<"
```

external_modules/total_energy_module/prepare_qe.sh
elcorto: I meant the `echo "HERE"` part which looks like a left over printf-debugging artifact.

external_modules/total_energy_module/prepare_qe.sh
elcorto: Side note, in the absence of `set -u`, I'd use `[ -z $foo ]` for undefined vars since any undefined shell variable `$foo` is a string of length 0. 

But to not further split hais here, I think the script might do what it should now.

external_modules/total_energy_module/prepare_qe.sh
RandomDefaultUser: Ok. great!

external_modules/total_energy_module/prepare_qe.sh
RandomDefaultUser: Aaaah now I understand what you mean. Yeah there is no reason to do it with a flag in this case, this was done more out of habit. I have edited the scripts to now require a path be provided and fail otherwise, could you have a final look?

elcorto: Looks good.

You can use `grep`'s exit code in the conditional, where succesful `grep` exits 0 (true), else exits 1 (false). Make sure to use single quotes, else the shell will try to expand out `$foxflags` in the script.

```sh
if grep -q '$foxflags -fPIC' install/configure; then
    ....
fi
```

RandomDefaultUser: I originally had that, but since there is nothing to do if grep exits 0 I instead opted for the comparison I currently have.

elcorto: Ah, sorry then I misread that. 

This would be acomplished by

```sh
if ! grep -q '$foxflags -fPIC'
```

RandomDefaultUser: Thanks, I tried that but it wouldn't work until I put 

```bash
if ! $(grep -q "\$foxflags -fPIC" install/configure);
```

I have pushed this, I think with this the scripts can be considered finished?

PR #350, opened 2022-10-07
Update descriptor.py for uneven processor count

dytnvgl:
This change allows a random number of parallel ranks to be used for calculating total energy in Quantum Espresso. This allows mala to write density unevenly to processors as expected by Quantum Espresso.


RandomDefaultUser:
Closed in favor of #354

PR #351, opened 2022-10-07
Update density.py to use uneven z splitting

dytnvgl:
When uneven zsplitting occurs each rank contains a nonuniform number of grid points. Quantum Espresso expects one value of gridpoints on all ranks. The padding of the density on lower grid point ranks allows Quantum Espresso to accept the density arrays as input for total energy calculations.


RandomDefaultUser:
Closed in favor of #354

Issue #3, opened 2021-03-31
OpenPMD integration


RandomDefaultUser:
In GitLab by @RandomDefaultUser on Dec 15, 2020, 10:02

File I/O should be changed to OpenPMD as soon as possible.

PR #357, opened 2022-11-02
Integrate OpenPMD into MALA - First round

RandomDefaultUser:
Update: Franz and me have been working a lot of this. 

### This PR includes:

- Full OpenPMD support - OpenPMD can now be used interchangeably to numpy
- A test for this interface
- Some under-the-hood data handling improvements; namely, the arrays holding data for ML runs are now allocated prior to loading, which saves on intermediate memory
- Refactoring in the Descriptor and Target classes (and subclasses) to accommodate the new, streamlined data procedures
- Density data is now always treated as either 4D or 2D data, which it should have been from the beginning

### This PR does NOT include: 

- Making OpenPMD the default framework - it does not come with a deprecation warning either, for the time being
- Making use of OpenPMD parallelization capabilities

Basic OpenPMD functionality (saving/loading) works. What's missing is performance improvements, testing and support for all targets.
Numpy interface has not been disabled, OpenPMD has simply been made the default. Numpy interface to be discontinued in MALA v1.3.0, so not the next release, but the one after.


mala/targets/ldos.py
```diff
@@ -520,6 +547,92 @@ def read_from_array(self, array, units="1/(eV*A^3)"):
...
+            # Since there is no sensible reason to not save new MALA data
+            # with the wrong units, for now we just check and complain.
+            # Eventually we may have to implement unit conversion, I guess.
+            if not np.isclose(ldos_mesh[str(i)].unit_SI, ((m**3)*J)):
+                raise Exception("MALA currently cannot operate with HDF5 "
+                                "files with non-MALA units.")
+
+            temp_ldos = ldos_mesh[str(i)].load_chunk()
+            series.flush()
+            self.local_density_of_states[:, :, :, i] = temp_ldos.copy()
```
franzpoeschel, line 606: I think that making this more efficient without error is a matter of picking the right variant of load_chunk(). 
My current hypothesis:
**What you wanted to do** Load data into an existing buffer
**What you accidentally did** Replace the existing buffer with the new buffer, created by `load_chunk`

There exists also a version of `load_chunk` that works with existing buffers, we'll have to carefully pick the right one here.

franzpoeschel (commented):
This looks generally good and straight-forward. I've left some comments on things that I noticed while skimming.
I'll start preparing some placeholder attributes tomorrow, so we have a template that we can then fill out together :)

mala/descriptors/descriptor.py
```diff
@@ -108,6 +148,133 @@ def backconvert_units(array, out_units):
...
+
+        # Check if this actually MALA compatible data.
+        if series.get_attribute("is_mala_data") != 1:
+            raise Exception("Non-MALA data detected, cannot work with this "
+                            "data.")
+
+        # A bit clanky, but this way only the FIRST iteration is loaded,
+        # which is what we need for loading from a single file that
+        # may be whatever iteration in its series.
+        for current_iteration_index in series.iterations:
```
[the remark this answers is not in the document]
franzpoeschel: (The same also goes at writing time, `series.write_iterations()`)

RandomDefaultUser:
Will close #3

setup.py
```diff
@@ -16,7 +16,7 @@
 
 extras = {
     'dev': ['bump2version'],
-    'opt': ['oapackage', 'pqkmeans'],
+    'opt': ['oapackage', 'pqkmeans', 'openpmd-api'],
```
DanielKotik, line 19: If a dependency (an optional one in this case) is added without updating the Conda environment accordingly at the same time, the test pipeline fails as it checks whether the provided environment meets all the specified requirements of MALA. 

In the CPU-test pipeline MALA gets installed with optional and test dependencies (as specified in  `setup.py`):

https://github.com/mala-project/mala/blob/2d8c5217ec83070a56c63b96a456296dfe9853f2/.github/workflows/cpu-tests.yml#L158

Since `openpmd` is not (yet) part of the Conda environment within the Docker container, it is simply installed additionally and thus the deviation from the given environment is detected.

image

As we want to provide an environment that fully meets the requirements of MALA, and also to avoid installation time of packages in CI jobs, such a deviation is detected and aborts the pipeline correctly. As a consequence this PR should not have been merged in the first place without fixing the Conda environment in the Docker container. This is now being resolved retrospectively in PR #383.

RandomDefaultUser: Understood! I will add a chapter in the documentation "contributing" section on how to properly handle the addition of such dependencies.

PR #359, opened 2022-11-15
Reduced memory footprint

RandomDefaultUser:
Scaling and loading is now done into a pre-allocated array.


RandomDefaultUser:
I'll be closing this here for now. The same changes have been applied in the context of the OpenPMD integration, albeit more thorough, and I expect a first merge there soon.

PR #381, opened 2022-12-20
Update workflows

DanielKotik:
Update workflow files to avoid using features and environments that will soon become deprecated.


RandomDefaultUser (approved):
LGTM!

PR #383, opened 2022-12-21
Fix openpmd installation in Docker

DanielKotik:


DanielKotik:
@RandomDefaultUser please fix the tests related to openpmd, see https://github.com/mala-project/mala/actions/runs/3750489354/jobs/6370410275#step:10:103

RandomDefaultUser (approved):
Thanks for fixing this @DanielKotik !

PR #390, opened 2022-12-23
fixed redundant learning rate rescaling on restart, now only rescales…

### LINKED ISSUE #374: Learning rate rescaled upon restarting.
From Kyle Miller: "I've noticed that upon restarting my trainings, the learning rate increases drastically each time due to the scaling in the __prepare_to_train() function in trainer.py. It seems like this behavior is meant to be limited to fresh models, rather than restarts."

kyledmiller:
… for fresh model

Require that last_epoch == 0 to do learning rate rescaling to avoid repeated rescaling when restarting training runs. Also had to move the rescaling code so it occurs after the last_epoch is initialized in the case of a restart.
resolves #374

### DIFF
--- a/mala/network/trainer.py
+++ b/mala/network/trainer.py
@@ -398,9 +398,13 @@ class Trainer(Runner):
         if self.parameters_full.use_gpu:
             kwargs['pin_memory'] = True
 
+        # Read last epoch
+        if optimizer_dict is not None: 
+            self.last_epoch = optimizer_dict['epoch']+1
+
         # Scale the learning rate according to horovod.
         if self.parameters_full.use_horovod:
-            if hvd.size() > 1:
+            if hvd.size() > 1 and self.last_epoch == 0:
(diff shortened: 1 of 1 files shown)


RandomDefaultUser (approved):
Looks good to me, I will merge this now. 
As a small side note, did you overwrite an earlier commit when implementing the changes discussed via e-mail? The commit history looks a bit off, since I only see one commit, but there should be two. Not that that's a problem per se, it might just be easier for future developments, if one can see the entire development history.

kyledmiller:
Yeah, I had actually made 5 commits testing various versions of the fix with github's built-in testing. Since 4 of them were just unsatisfactory tests, I squashed them since typically I would do that testing and settle on a working fix before committing. I can leave the full history in the future though.

RandomDefaultUser:
Ah sorry, you are right and I am wrong. Of course squashing commits the way you do is a good idea when testing things, I was confused by how Github potrayed that. So feel free to continue as you did :)

PR #392, opened 2022-12-25
Update environment

DanielKotik:
This PR updates the environment in various respects:

- Use latest Miniconda version in the `Dockerfile`
- Omit the build info from the environment definition export via `--no-builds` flag, i.e. `conda env export -n mala-cpu --no-builds`. These platform-specific build constraints can cause problems. This restriction isn't absolutely necessary for reproducibility rather it's sufficient to specify the exact package version numbers (the build information can still be retrieved from the Docker Conda environment if required).
- Fix missing `pandas` dependency (see https://github.com/mala-project/mala/commit/d3a15ea6425f4ec7b06dda57228ee1668316d4be#commitcomment-94234580)
- Problems installing optional package `pqkmeans`: had to manually downgrade `setuptools` (65.3.0 → 59.8.0)


RandomDefaultUser (approved):
Looks good to me!

PR #393, opened 2022-12-28
Fix mirror workflow

DanielKotik:
This PR fixes the mirror workflow. This was broken for months. 

With this PR I also wanted to enable host verification:

> To obtain the public key for the `GIT_SSH_KNOWN_HOSTS` secret, we have
    issued the command `ssh-keyscan -t rsa github.com`.

However, this fails and will probably be fixed once [this PR](https://github.com/yesolutions/mirror-action/pull/19) of the mirror action used is merged.


RandomDefaultUser (approved):
Looks good to me!

Issue #60, opened 2021-03-31
Runtime warning/Overflow in test-workflow


RandomDefaultUser:
In GitLab by @RandomDefaultUser on Mar 22, 2021, 11:14

The following part of the `test-workflow` log shows an overflow:
```
234  from .density import Density
235  /builds/multiscale-wdm/surrogate-models/fesl/fesl/fesl/targets/calculation_helpers.py:143: RuntimeWarning: overflow encountered in exp
236  return 1.0 / (1.0 + np.exp((energy_ev - fermi_energy_ev) /
```

elcorto:
Ran tests and examples recently (#273, #274), and this is still there.

PR #400, opened 2023-01-06
Added overflow check in Fermi function

RandomDefaultUser:
This adds a check to make sure that the calculation of exp(x) does not result in an overflow in the Fermi function. It gets rid of the Runtime warning encountered in #60 .

### DIFF
--- a/mala/targets/calculation_helpers.py
+++ b/mala/targets/calculation_helpers.py
@@ -3,7 +3,7 @@ from ase.units import kB
 import mpmath as mp
 import numpy as np
 from scipy import integrate
-
+import sys
 
 def integrate_values_on_spacing(values, spacing, method, axis=0):
     """
@@ -72,9 +72,13 @@ def fermi_function(energy, fermi_energy, temperature):
     # therefore get zero (or machine precision within zero) either way.
     # Adding this check removes overflow warnings.
     # The overhead for inference should be minimal.
-    max_exponent = np.finfo(exponent.dtype.name).max
-    overflow = np.argwhere(exponent > max_exponent)
-    exponent[overflow] = max_exponent
+    # Since this function works both for arrays and scalars,
+    # we have to check which maximum to use.
+    if type(energy) == np.ndarray:
+        max_exponent = np.log(np.finfo(exponent.dtype.name).max)
+        exponent[np.argwhere(exponent > max_exponent)] = max_exponent
+    else:
+        exponent = min(exponent, np.log(sys.float_info.max))
 
     return 1.0 / (1.0 + np.exp(exponent))
 
--- a/mala/targets/calculation_helpers.py
+++ b/mala/targets/calculation_helpers.py
@@ -74,11 +74,11 @@ def fermi_function(energy, fermi_energy, temperature):
     # The overhead for inference should be minimal.
     # Since this function works both for arrays and scalars,
     # we have to check which maximum to use.
-    if type(energy) == np.ndarray:
-        max_exponent = np.log(np.finfo(exponent.dtype.name).max)
-        exponent[np.argwhere(exponent > max_exponent)] = max_exponent
+    if isinstance(energy, np.ndarray):
+        max_exponent = np.log(np.finfo(exponent.dtype).max)
+        exponent[exponent > max_exponent] = max_exponent
     else:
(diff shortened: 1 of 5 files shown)


mala/targets/calculation_helpers.py
```diff
@@ -62,8 +62,25 @@ def fermi_function(energy, fermi_energy, temperature):
...
+    # np.exp() (for double values only).
+    # Very large exponents result in very large values, and if we don't check
+    # for overflow, infinity. Cutting this back to the maximum float value
+    # (~ 1e308) is more then enough, since we divide by the exponent, and
+    # therefore get zero (or machine precision within zero) either way.
+    # Adding this check removes overflow warnings.
+    # The overhead for inference should be minimal.
+    # Since this function works both for arrays and scalars,
+    # we have to check which maximum to use.
+    if type(energy) == np.ndarray:
```
elcorto, line 77: `isinstance(energy, np.ndarray)` is the more general way to do this (even though we won't expect to get passed ndarray subclasses here).

mala/targets/calculation_helpers.py
```diff
@@ -62,8 +62,25 @@ def fermi_function(energy, fermi_energy, temperature):
...
+        max_exponent = np.log(np.finfo(exponent.dtype.name).max)
```
elcorto, line 78: `dtype.name` is not necessary, `finfo(exponent.dtype)` does it.

mala/targets/calculation_helpers.py
```diff
@@ -62,8 +62,25 @@ def fermi_function(energy, fermi_energy, temperature):
...
+        exponent[np.argwhere(exponent > max_exponent)] = max_exponent
```
elcorto, line 79: Hmm, is `exponent` 1d, so shape `(N,)` ? `argwhere adds an extra dimension, not sure of that's what you want.

```py
>>> a=np.random.rand(100)
>>> a[np.argwhere(a>0.5)].shape
(51, 1)
>>> a[a>0.5].shape
(51,)
```

Why not use the bood mask `exponent > max_exponent` directly?

mala/targets/calculation_helpers.py
```diff
@@ -62,8 +62,25 @@ def fermi_function(energy, fermi_energy, temperature):
...
+    else:
+        exponent = min(exponent, np.log(sys.float_info.max))
```
elcorto, line 81: Here you assume that `exponent` is `float64` (since this is what Python's built-in float defaults to and what `sys.float_info` talks about).

```py
>>> type(1.0)
float
```

Note that `np.finfo` also accepts numbers, so all these are the same

```py
>>> np.finfo(1.0)
>>> np.finfo(type(1.0))
>>> np.finfo("float")
>>> np.finfo("float64")
```

As such, you don't need `sys.float_info` and can instead query `exponent` using it's actual type: `finfo(type(exponent))`.

elcorto (commented):
The ideal solution would be to fix the reason for the overflow, which is unrealistic energy values as stated in #60, rather than silencing warnings. But in the case here setting a max exponent is ok since, as you mention, the result will be close to zero anyway.

However, I suggest to not make this the default behavior. One could have a flag such as `clip_exponent` which is False by default and will emit the overflow RuntimeWarning in any case where the function is used outside the specific use case mentioned in #60 and where such as warning may give hints to other issues. In the specific context from #60, one would then call `fermi_function(..., clip_exponent=True)` to explicitly silence the warning. Since you have much more overview of the code base, this may be skipped when `fermi_function()` is only ever used in the #60 use case.

mala/targets/calculation_helpers.py
RandomDefaultUser: Got it, thanks!

mala/targets/calculation_helpers.py
RandomDefaultUser: Ah yeah, that is much better! Implemented in 3f7bf58fa9f2a7544df45eecb158ce63f954dee1.

mala/targets/calculation_helpers.py
RandomDefaultUser: Oh I didn't know that, implemented it in 3f7bf58fa9f2a7544df45eecb158ce63f954dee1.

PR #404, opened 2023-01-11
Flexible data handling: allow different grid sizes

timcallow:
This PR removes the requirement to have snapshots of fixed grid sizes (unless clustering + lazy loading is used, when fixed grid sizes are still required).

Attached is a minimal example to see that it works for snapshots of different grid sizes.

Additional notes:
- This is basically a merge of [this commit](https://github.com/mala-project/mala/commit/60e14a82bb6601b7ac2078f2886b439abaf01d95) and the dev branch
- Where I wasn't sure if something I added was coded in the best way, I've tried to add a TODO statement
- Running tests locally, `inference_test.py` has some failures (see attached log). Not sure why, still working on it, maybe @RandomDefaultUser knows? Other tests seem ok

[inference_test.log](https://github.com/mala-project/mala/files/10394651/inference_test.log)
[data_handling_test.tar.gz](https://github.com/mala-project/mala/files/10394592/data_handling_test.tar.gz)


mala/datahandling/data_handler.py
```diff
@@ -380,6 +386,8 @@ def raw_numpy_to_converted_scaled_tensor(self, numpy_array, data_type,
         converted_tensor: torch.Tensor
             The fully converted and scaled tensor.
         """
+        # TODO: not sure where this function is used
+        # since we got rid of the self.grid_size parameter, it won't work in current form
```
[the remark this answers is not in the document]
RandomDefaultUser: This function is also needed in the CI, and since it is currently broken, the test pipeline fails.

mala/datahandling/data_handler.py
```diff
@@ -478,21 +486,23 @@ def __check_snapshots(self):
...
+            # for flexible grid sizes only this need be consistent
             tmp_input_dimension = tmp_dimension[-1]
             tmp_grid_dim = tmp_dimension[0:3]
+            # TODO: move assignment of snapshot dims to snapshot.py??            
+            snapshot.grid_dimension = tmp_grid_dim
+            snapshot.grid_size = int(np.prod(snapshot.grid_dimension))            
             if firstsnapshot:
                 self.input_dimension = tmp_input_dimension
-                self.grid_dimension[0:3] = tmp_grid_dim[0:3]
+                # TODO is this the best way to do this?
```
RandomDefaultUser, line 498: See comment above, I think for now it is, but we have to open an issue.

mala/datahandling/data_handler.py
```diff
@@ -515,78 +525,67 @@ def __check_snapshots(self):
...
+                if self.nr_validation_snapshots != 0:
+                    printout("As this DataHandler can only be used for "
+                             "inference, the validation data you have "
+                             "provided will be ignored.", min_verbosity=1)
+            if self.nr_test_snapshots == 0:
+                printout("Running MALA without test data. If this is not "
+                         "what you wanted, "
+                         "please revise the input script.", min_verbosity=0)
 
-            # MALA can either be run in training or test-only mode.
```
RandomDefaultUser, line 561: I simplified this comparison in a recent PR, I think it got reverted by accident in this PR.

mala/datahandling/data_handler.py
```diff
@@ -695,14 +666,16 @@ def __load_data(self, function, data_type):
                     units = snapshot.output_units
 
                 if snapshot.snapshot_type == "numpy":
-                    calculator.read_from_numpy_file(file, units=units,
-                                                    array=getattr(self, array)[snapshot_counter])
+                    tmp_array = calculator.read_from_numpy_file(file, units=units)
+                    tmp_array = tmp_array.reshape([gs_new, feature_dimension])
+                    getattr(self, array)[gs_old : gs_old + gs_new] = tmp_array
```
RandomDefaultUser, line 671: The idea of giving the array directly to the function is to save on intermediate memory. By assigning a temporary array, we always have to have extra memory for at least one temporary array. 

What I would suggest is modifying the `read_from_numpy_file` function by adding a reshape option there. Then the same code would work again. Could you try that or else let me know so I will give it a try?

Also could you add the reshape to the OpenPMD section? There we have to bite the bullet with the temporary arrays anyhow because of the way we use OpenPMD.

mala/datahandling/snapshot.py
```diff
@@ -73,7 +76,65 @@ def __init__(self, input_npy_file, input_npy_directory,
...
-        self.snapshot_type = snapshot_type
+        self.snapshot_type = snapshot_type        
+
+        # All the dimensionalities of the snapshot.
+        self.grid_dimensions = None
+        self.grid_size = None
+        self.input_dimension = None
+        self.output_dimension = None
+        
+    def load_dimensions(self, descriptors_contain_xyz,
```
RandomDefaultUser, line 87: This function should be obsolete now. The PhysicalData class (which all targets and descriptors inherit from) have `load_dimensions` functions for both OpenPMD and numpy and I would like to keep using those since we then only have to maintain code at one location. 

This probably requires some changes in the DataHandler again, as the snapshot locations have to be passed to the calculators. Sorry for the mess, I added this function (I think) before the PhysicalData class was introduced.

test/examples_test.py
```diff
@@ -34,6 +34,7 @@ def test_ex08(self):
         runpy.run_path("../examples/ex08_checkpoint_hyperopt.py")
 
     def test_ex09(self):
+
```
RandomDefaultUser, line 37: I think an empty line was erronously commited here.

RandomDefaultUser (changes requested):
Thanks a ton @timcallow for this PR! I know this is a lot of work, since it requires wrestling with some nasty MALA internals, so I appreciate it. 
Most of it looks good. I have commented some changes that I think need to be made, as well as the cause for the failing tests. Let me know if you need help fixing some of the more obscure features.

mala/datahandling/data_handler.py
timcallow: Should be in sync with dev now!

mala/datahandling/data_handler.py
```diff
@@ -66,24 +66,28 @@ def __init__(self, parameters: Parameters, target_calculator=None,
...
-        self.grid_dimension = [0, 0, 0]
-        self.grid_size = 0
+        self.input_dimension = 0 # None ??
+        self.output_dimension = 0 # None ??
         self.nr_snapshots = 0
+        
+        # TODO only define these parameters in case of clustering
+        if self.parameters.use_clustering:
+            self.grid_dimension = [0, 0, 0]
+            self.grid_size = 0
```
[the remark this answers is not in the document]
timcallow: I've opened issue #408 to document this

RandomDefaultUser (approved):
This looks good to me now. Thanks for all the hard work @timcallow ! I will merge this now.

PR #405, opened 2023-01-12
Accelerating TEM with GPU

### LINKED ISSUE #362: Port TEM to GPU
For more efficient inference, it would be advantageous to have a GPU accelerated version of the TEM. 
This include steps:

1. Port `external_modules/total_energy_module/total_energy.f90` to a GPU ready version of QE.
2. Test that this works as expected (e.g. with `examples/ex03_postprocess_data.py`)
3. Test whether the functions we lose time in are already GPU ready; they may not be, since we run at different scales then QE regularly does.

RandomDefaultUser:
This PR will allow the calculations in the TEM to be accelerated with GPU. Will close #362 .


external_modules/total_energy_module/build_total_energy_module.sh
```diff
@@ -38,9 +38,16 @@ modobjs="$(find $root_dir/Modules/ -name "*.o")"
...
 
-project_lib_folders=" -L$root_dir/Modules -L$root_dir/KS_Solvers -L$root_dir/FFTXlib -L$root_dir/LAXlib -L$root_dir/UtilXlib -L$root_dir/dft-d3"
-project_libs="-lqemod -lks_solvers -lqefft -lqela -lutil -ldftd3qe"
-project_inc_folders="-I$root_dir/Modules -I$root_dir/FFTXlib -I$root_dir/LAXlib -I$root_dir/KS_Solvers -I$root_dir/UtilXlib"
+# LF: For some reason I cannot make the linker find the XClib if I don't rename
+# the library. This is probably just me not understanding linker syntax well
+# enough, but after this rename it works. If someone has a smarter idea,
+# I am all ears.
+
+cp $root_dir/XClib/xc_lib.a $root_dir/XClib/libqexc.a
```
RandomDefaultUser, line 46: I know that this solution is hacky, but I couldn't find a better. Since you wrote most of the script @elcorto , do you maybe have an idea? The problem is that with the original name of the XClib library file, I cannot make the linker find it, so I renamed it to adhere to the standard naming scheme.

elcorto: So the linker doesn't pick it up if it is not named `lib*.a`? Since it is a static lib you should be able to pass it directly as `$root_dir/XClib/xc_lib.a` to `f2py` as we did with `$root_dir/clib/clib.a` in the old version of the script.

RandomDefaultUser: Yes, exactly, the linker doesn't pick it up properly. I will try your suggestion!

RandomDefaultUser: Tried it, it worked. Thanks!

PR #409, opened 2023-01-17
Incorporated changes from Josh Romero (Nvidia)

RandomDefaultUser:
All actual code edits by Josh Romero (@romerojosh) from Nvidia. I integrated the performance improvements he coded up into the most recent version of MALA. I omitted some parts of the code that seemed to have been left over from the testing phase. I hope I didn't break anything, and I am currently testing to confirm I did not.

### DIFF
--- a/mala/common/parameters.py
+++ b/mala/common/parameters.py
@@ -574,6 +574,10 @@ class ParametersData(ParametersBase):
     sample_ratio : float
         If use_clustering is True, this is the ratio of training data used
         for sampling per snapshot (according to clustering then, of course).
+
+    use_fast_tensor_data_set : bool
+        If True, then the new, fast TensorDataSet implemented by Josh Romero
+        will be used.
     """
 
     def __init__(self):
@@ -588,6 +592,8 @@ class ParametersData(ParametersBase):
         self.number_of_clusters = 40
         self.train_ratio = 0.1
         self.sample_ratio = 0.5
+        self.use_fast_tensor_data_set = False
+
 
 class ParametersRunning(ParametersBase):
     """
@@ -687,6 +693,21 @@ class ParametersRunning(ParametersBase):
     inference_data_grid : list
         List holding the grid to be used for inference in the form of
         [x,y,z].
+
+    use_mixed_precision : bool
+        If True, mixed precision computation (via AMP) will be used.
+
+    use_graphs : bool
+        If True, then CUDA graphs are used during training for improved
+        performance.
+
+    training_report_frequency : int
+        Determines how often detailed performance info is printed during
+        training (only has an effect if the verbosity is high enough).
+
+    profiler_range : list
+        List with two entries determining with which batch size the CUDA
+        profiler will start and stop profiling.
     """
 
     def __init__(self):
@@ -713,6 +734,10 @@ class ParametersRunning(ParametersBase):
         self.during_training_metric = "ldos"
         self.after_before_training_metric = "ldos"
         self.inference_data_grid = [0, 0, 0]
+        self.use_mixed_precision = False
+        self.use_graphs = False
+        self.training_report_frequency = 1000
+        self.profiler_range = [1000, 2000]
 
     def _update_horovod(self, new_horovod):
         super(ParametersRunning, self)._update_horovod(new_horovod)
--- a/mala/network/trainer.py
+++ b/mala/network/trainer.py
@@ -2,6 +2,7 @@
 import os
 import time
 from datetime import datetime
+from packaging import version
 
 try:
     import horovod.torch as hvd
@@ -18,6 +19,7 @@ from mala.common.parameters import Parameters
 from mala.common.parameters import printout
 from mala.datahandling.data_handler import DataHandler
 from mala.datahandling.data_scaler import DataScaler
+from mala.datahandling.fast_tensor_dataset import FastTensorDataset
 from mala.network.network import Network
 from mala.network.runner import Runner
 
@@ -83,6 +85,14 @@ class Trainer(Runner):
             self.tensor_board = SummaryWriter(self.full_visualization_path)
             printout("Writing visualization output to",
                      self.full_visualization_path, min_verbosity=1)
+
+        self.gradscaler = None
+        if self.parameters.use_mixed_precision:
+            printout("Using mixed precision via AMP.", min_verbosity=1)
+            self.gradscaler = torch.cuda.amp.GradScaler()
+
+        self.train_graph = None
+        self.validation_graph = None
 
     @classmethod
     def checkpoint_exists(cls, checkpoint_name, use_pkl_checkpoints=False):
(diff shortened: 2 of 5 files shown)


RandomDefaultUser:
benchmarking

I benchmarked these changes on the Al256@RT dataset to confirm I could reproduce the speed ups, and as you can see, I easily could. I also benchmarked the inference performance of the resulting models. Please note that these models have only been trained for 50 epochs, so the performance is a bit worse than the one reported in the paper. 
At any rate, the performance is pretty good, the only drop in performance I see is from using `FusedAdam` - is that something that is known?
Apart from that I think this PR can be merged very soon. 
To-Dos before merging from my side would be:

- [ ] Document that CUDA graphs are only available with CUDA 11.0 and above, so that users do not get crashes by accident
- [ ] Write a doc page detailing how all these performance improvements can be used in MALA and decide which maybe should be the default
- [ ] Refactor the trainer class a bit - since a lot of code was added there it is now a tad long, maybe I can refactor some parts into their own routines to make it more accessible
- [ ] Resolve Merge conflicts
- [ ] Wait on final feedback from Josh

After merging I would like to tackle the lazy loading process in a similar fashion. The improvements seen here are wonderful, and I would love to apply those to lazy loading as well. Most of it should be there by default, with the exception of the FastTensorDataset - that is maybe something we could do.

RandomDefaultUser:
I fixed the remaining items. I think we only need to final feedback from Josh/Sandia side and then we can merge.

srajama1:
@romerojosh : Thanks for the changes, we are seeing the improvements as well. Thanks, @RandomDefaultUser ! 

It is interesting the errors went up with FusedADAM. @romerojosh : Have you seen this as well. Can we check there are no subtle issues here. The performance improvement there is not much higher if we add more epochs to get that back down, right?

romerojosh:
Very nice to see these changes getting committed and glad to see that folks are able to reproduce the performance improvements with the changes. I will review this as soon as I can.

@srajama1 I wouldn't expect there to be an issue with FusedAdam since it should be numerically equivalent to the native torch Adam. I didn't see this in my testing, but I also didn't have a "golden" reference result to compare to. From the plots above, it looks like the MAE values are all over the place; comparing MAE from the baseline to the Opt3 bar shows a difference of about of -7 while the difference from baseline to the Opt4 bar with FusedAdam is +10. What is the expected value here? Is it just that none of these results are quite converged?

docs/source/usage/trainingmodel.rst
```diff
@@ -5,3 +5,43 @@ Neural networks are powerful machine learning tools in principle capable of
...
+        """
+        This setting is pretty straightforward - wherever possible, use a GPU.
+        """
+        parameters.use_gpu = False # True: Use GPU
+        """
+        Multiple workers allow for faster data processing, but require
+        additional CPU/RAM power. A good setup is e.g. using 4 CPUs attached
+        to one GPU and setting the num_workers to 4.
+        Please be aware that using multiple workers is currently not supported
+        with horovod.
```
romerojosh, line 29: Is the limitation for multiple workers and horovod something that is unique to the `FastTensorDataset` introduced or was it already some existing limitation in MALA?

docs/source/usage/trainingmodel.rst
```diff
@@ -5,3 +5,43 @@ Neural networks are powerful machine learning tools in principle capable of
...
+        parameters.running.num_workers = 0 # set to e.g. 4
+        """
+        MALA supports a faster implementation of the TensorDataSet class
+        from the torch library. Turning it on will drastically improve
+        performance.
+        """
+        parameters.data.use_fast_tensor_data_set = False # True: Faster data loading
+        """
+        Likewise, using CUDA graphs improve performance by optimizing GPU
+        usage. Be careful, this option is only availabe from CUDA 11.0 onwards.
```
romerojosh, line 40: It may be a good idea to also mention that CUDA graphs will be most effective in cases that are latency-limited (e.g. small models with very short training iterations). The reason for this is that CUDA graphs alleviate many of the efficiency losses due to PyTorch kernel launch overheads, which are more prominent when running small models.

mala/common/parameters.py
```diff
@@ -687,6 +693,17 @@ class ParametersRunning(ParametersBase):
...
+
+    use_mixed_precision : bool
+        If True, mixed precision computation (via AMP) will be used.
+
+    training_report_frequency : int
+        Determines how often detailed performance info is printed during
+        training (only has an effect if the verbosity is high enough).
+
+    profiler_range : list
+        List with two entries determining with which batch size the CUDA
```
romerojosh, line 705: I think this should be "iteration range" instead of "batch size" in the documentation here. Might also want to mention that this option is only relevant when running using the `nsys` profiler with the `--capture-range cudaProfilerApi` option enabled.

mala/network/trainer.py
```diff
@@ -412,15 +486,27 @@ def __prepare_to_train(self, optimizer_dict):
...
+                                      weight_decay=self.parameters.
+                                      weight_decay)
         elif self.parameters.trainingtype == "Adam":
+            printout("Using torch.optim.Adam optimizer...", min_verbosity=2)
             self.optimizer = optim.Adam(self.network.parameters(),
                                         lr=self.parameters.learning_rate,
                                         weight_decay=self.parameters.
                                         weight_decay)
+        elif self.parameters.trainingtype == "FusedAdam":
+            printout("Using Apex FusedAdam optimizer...", min_verbosity=2)
```
romerojosh, line 501: I see that instead of the `FusedAdam` from the `apex` library, you've opted to use the new `fused` option in the native `torch.optim.Adam` optimizer, which is a good change. The printout here should be updated to reflect that though.

romerojosh (commented):
Adding just a few initial comments after a first pass through the changes. I will review again and approve once I test the PR out on my system (and the MAE results from the earlier comments are worked out if necessary).

RandomDefaultUser:
> Very nice to see these changes getting committed and glad to see that folks are able to reproduce the performance improvements with the changes. I will review this as soon as I can.
> 
> @srajama1 I wouldn't expect there to be an issue with FusedAdam since it should be numerically equivalent to the native torch Adam. I didn't see this in my testing, but I also didn't have a "golden" reference result to compare to. From the plots above, it looks like the MAE values are all over the place; comparing MAE from the baseline to the Opt3 bar shows a difference of about of -7 while the difference from baseline to the Opt4 bar with FusedAdam is +10. What is the expected value here? Is it just that none of these results are quite converged?

@romerojosh You are right, none of these results are quite converged. I only trained for 50 epochs each, convergence may take a bit longer. The results should be below 10 meV/atom, which is almost universally the case except for `FusedAdam`, hence my question. But it may just be a bad network init. I will train 5  networks per optimization level and report back!

docs/source/usage/trainingmodel.rst
RandomDefaultUser: This predates the `FastTensorDataset` and is something that @ellisja had told me to look out for. If I recall correctly, the problem is that in order for multiple workers to work correctly, we need more CPUs, but that would lead to horovod launching more ranks? I guess I have not tested it by e.g. requesting 8 CPUs from slurm but then only launching 2 MPI jobs... To be fully honest, I avoid using horovod at the moment because it has given me a lot of problems with running larger jobs, so I think this limitation would not be a problem for me.

mala/common/parameters.py
RandomDefaultUser: Oh yes, you are right, I worded that weird.

docs/source/usage/trainingmodel.rst
RandomDefaultUser: Done!

mala/network/trainer.py
RandomDefaultUser: Good point! On second thought, I removed all the printouts - they shouldn't be necessary, since choosing the optimizer is done via the parameters and therefore accesible via there.

RandomDefaultUser:
> Adding just a few initial comments after a first pass through the changes. I will review again and approve once I test the PR out on my system (and the MAE results from the earlier comments are worked out if necessary).

Thanks for the feedback, I made the necessary changes! Let me know if there are further things I should adapt. The accuracy problem is most likely just due to a faulty metric, and I'll report back once I have run the necessary tests.

docs/source/usage/trainingmodel.rst
romerojosh: From a GPU workload perspective, you would typically launch a single process per GPU and then set `num_workers` to a value up to the number of cores you have available per process. For example, if you had a 32 core CPU and 4 GPUs, you might consider launching 4 MPI ranks per node, you could set `num_workers` up to 8 and not oversubscribe CPU resources (though in reality, you'd set it less than 8 to leave room for PyTorch and other threads). 

I wonder if the issue that was brought up in the past had to do with multiple workers using `fork` to create the worker processes and that causing hangs when using Horovod. This is due to an issue in some MPI implementations having issues (e.g. hanging) when a process forks. This can usually be worked around by having PyTorch use `spawn` or `forkserver` to launch processes instead (see https://github.com/horovod/horovod/blob/85619c6851ef4ae3e514e9351bae8a9ca33c2d3b/examples/pytorch/pytorch_imagenet_resnet50.py#L211-L215 as an example)

RandomDefaultUser: I totally see your point and that sounds reasonable. I think that should already work with MALA if the MPI configuration is correct, I only haven't tried it out.

RandomDefaultUser:
I reran the models, with 5 models per optimization level, and averaged the MAE over those; the black bars indicate the standard deviation. 
benchmarking


The accuracy is stable across the board. `FusedAdam` is peforming the "worst" in both terms of MAE and spread - but by a very, very small amount. This may still be just noise, and even if not, the change is asbolutely negligible. 

Thanks for reviewing and approving @romerojosh , I will merge now :)

srajama1:
@RandomDefaultUser Thanks for checking the accuracy and confirming everything is good.

Issue #433, opened 2023-03-28
Accuracy wildly wrong when loading numpy32 data


RandomDefaultUser:
For some reason, loading numpy32 data gives wildly wrong results for LDOS integration. This seems like a bug, since NN internally use FP32.

PR #434, opened 2023-03-28
Implement lazy loading changes from Josh Romero

RandomDefaultUser:
Implements the lazy loading prefetching changes from @romerojosh into the MALA code base. Currently cannot test fully due to #433.

### DIFF
--- a/mala/datahandling/multi_lazy_load_data_loader.py
+++ b/mala/datahandling/multi_lazy_load_data_loader.py
@@ -1,0 +1,117 @@
+import os
+
+import numpy as np
+from torch.utils.data import DataLoader
+
+from multiprocessing import shared_memory
+import concurrent.futures
+
+# Class to handle loading multiple LazyLoadDatasetSingle objects with
+# prefetching
+
+
+class MultiLazyLoadDataLoader:
+    def __init__(self, datasets, **kwargs):
+        self.datasets = datasets
+        self.loaders = []
+        for d in datasets:
+            self.loaders.append(DataLoader(d,
+                                          batch_size=None,
+                                          **kwargs,
+                                          shuffle=False))
+
+        # Create single process pool for prefetching
+        # Can use ThreadPoolExecutor for debugging.
+        #self.pool = concurrent.futures.ThreadPoolExecutor(1)
+        self.pool = concurrent.futures.ProcessPoolExecutor(1)
+
+        # Allocate shared memory and commence file load for first
+        # dataset in list
+        dset = self.datasets[0]
+        dset.allocate_shared_mem()
+        self.load_future = self.pool.submit(self.load_snapshot_to_shm,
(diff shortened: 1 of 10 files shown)


RandomDefaultUser:
When running, I get the following warnings at the end of the run:

```
Exception ignored in: <function _ExecutorManagerThread.__init__.<locals>.weakref_cb at 0x7f0ac9792290>
Traceback (most recent call last):
  File "/usr/lib/python3.10/concurrent/futures/process.py", line 285, in weakref_cb
AttributeError: 'NoneType' object has no attribute 'util'
Exception ignored in: <function _ExecutorManagerThread.__init__.<locals>.weakref_cb at 0x7f0ac9790b80>
Traceback (most recent call last):
  File "/usr/lib/python3.10/concurrent/futures/process.py", line 285, in weakref_cb
AttributeError: 'NoneType' object has no attribute 'util'
 /usr/lib/python3.10/multiprocessing/resource_tracker.py:224: UserWarning: resource_tracker: There appear to be 4 leaked shared_memory objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

Nothing else breaks. The first error seems to be related to https://github.com/SergeyPirogov/webdriver_manager/issues/429#issuecomment-1426203643
The second I cannot make sense of. I added a cleanup function, but the error persists.

mala/datahandling/multi_lazy_load_data_loader.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # TODO: Without this function, I get 2 times the number of snapshots
+    # memory leaks after shutdown. With it, I get 1 times the number of
+    # snapshots. So using this function seems to be correct; just simply not
+    # enough? I am not sure where the memory leak is coming from.
+    def cleanup(self):
+        """Deallocate arrays still left in memory."""
+        for dset in self.datasets:
+            dset.delete_data()
+
```
romerojosh, line 115: In trying to deal with the leaked `shared_memory` warnings at job termination, I added:
```
self.pool.shutdown()
```
to stop the process pool before shutdown in this `cleanup` method. This didn't resolve the warnings, but is something that should be added anyway.

I wonder if we are running into the Python bug mentioned in this StackOverflow post?: https://stackoverflow.com/questions/62748654/python-3-8-shared-memory-resource-tracker-producing-unexpected-warnings-at-appli

RandomDefaultUser: I just tested this and it removed two of the three warnings (the weakref ones) for me, thanks a lot! :) 

As for the Stack Overflow post: I found that too and think you might be right, that could be related...

RandomDefaultUser: Fixed as per @romerojosh suggestions. The PR is ready to be merged now.

PR #435, opened 2023-03-29
Save and load training data in FP32

### LINKED ISSUE #433: Accuracy wildly wrong when loading numpy32 data

RandomDefaultUser:
Since pytorch operates with FP32 anyway, it would be a good idea to have the data saved in FP32 too. This seemed like a problem since LDOS integration seemed to go wrong when performed with FP32 data, but this was an oversight in the usage of the `np.sum` function. This PR fixes this, closes #433 and allows for native training data conversion into FP32. FP64 conversion is supported as legacy option.

### DIFF
--- a/mala/descriptors/bispectrum.py
+++ b/mala/descriptors/bispectrum.py
@@ -175,12 +175,6 @@ class Bispectrum(Descriptor):
 
             # Copy the grid dimensions only at the end.
             self.grid_dimensions = [nx, ny, nz]
-
-            # If I directly return the descriptors, this sometimes leads
-            # to errors, because presumably the python garbage collection
-            # deallocates memory too quickly. This copy is more memory
-            # hungry, and we might have to tackle this later on, but
-            # for now it works.
             return snap_descriptors_np, nrows_local
 
         else:
@@ -194,14 +188,6 @@ class Bispectrum(Descriptor):
             snap_descriptors_np = snap_descriptors_np.transpose([2, 1, 0, 3])
             # Copy the grid dimensions only at the end.
             self.grid_dimensions = [nx, ny, nz]
-            # If I directly return the descriptors, this sometimes leads
-            # to errors, because presumably the python garbage collection
-            # deallocates memory too quickly. This copy is more memory
-            # hungry, and we might have to tackle this later on, but
-            # for now it works.
-            # I thought the transpose would take care of that, but apparently
-            # it does not necessarily do that - so we have do go down
-            # that route.
             if self.parameters.descriptors_contain_xyz:
                 return snap_descriptors_np, nx*ny*nz
             else:
--- a/mala/descriptors/bispectrum.py
+++ b/mala/descriptors/bispectrum.py
@@ -93,6 +93,11 @@ class Bispectrum(Descriptor):
     def _calculate(self, atoms, outdir, grid_dimensions, **kwargs):
         """Perform actual bispectrum calculation."""
         from lammps import lammps
+        if "use_fp64" in kwargs.keys():
+            use_fp64 = kwargs["use_fp64"]
+        else:
+            use_fp64 = False
+
         lammps_format = "lammps-data"
         ase_out_path = os.path.join(outdir, "lammps_input.tmp")
         ase.io.write(ase_out_path, atoms, format=lammps_format)
@@ -165,7 +170,8 @@ class Bispectrum(Descriptor):
             snap_descriptors_np = \
                 extract_compute_np(lmp, "bgridlocal",
                                    lammps_constants.LMP_STYLE_LOCAL, 2,
-                                   array_shape=(nrows_local, ncols_local))
+                                   array_shape=(nrows_local, ncols_local),
+                                   use_fp64=use_fp64)
 
             # Copy the grid dimensions only at the end.
             self.grid_dimensions = [nx, ny, nz]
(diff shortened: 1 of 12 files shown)


mala/descriptors/bispectrum.py
```diff
@@ -175,13 +181,14 @@ def _calculate(self, atoms, outdir, grid_dimensions, **kwargs):
             # deallocates memory too quickly. This copy is more memory
             # hungry, and we might have to tackle this later on, but
             # for now it works.
-            return snap_descriptors_np.copy(), nrows_local
+            return snap_descriptors_np, nrows_local
```
elcorto, line 184: The comment above says that the `copy()` is needed, but this PR removes it.

mala/descriptors/bispectrum.py
```diff
@@ -196,6 +203,6 @@ def _calculate(self, atoms, outdir, grid_dimensions, **kwargs):
             # it does not necessarily do that - so we have do go down
             # that route.
             if self.parameters.descriptors_contain_xyz:
-                return snap_descriptors_np.copy(), nx*ny*nz
+                return snap_descriptors_np, nx*ny*nz
             else:
-                return snap_descriptors_np[:, :, :, 3:].copy(), nx*ny*nz
```
elcorto, line 201: The comment above says that the copy() is needed, but this PR removes it.

mala/descriptors/bispectrum.py
```diff
@@ -196,6 +203,6 @@ def _calculate(self, atoms, outdir, grid_dimensions, **kwargs):
             # it does not necessarily do that - so we have do go down
             # that route.
             if self.parameters.descriptors_contain_xyz:
-                return snap_descriptors_np.copy(), nx*ny*nz
```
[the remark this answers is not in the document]
RandomDefaultUser: The copy has been moved in the `lammps_util.py` file, but you are right, the comment should be adapted. Will do that.

mala/descriptors/bispectrum.py
RandomDefaultUser: The copy has been moved in the lammps_util.py file, but you are right, the comment should be adapted. Will do that.

mala/descriptors/bispectrum.py
RandomDefaultUser: The copy has been moved in the lammps_util.py file, but you are right, the comment should be adapted. Will do that.

RandomDefaultUser:
> General question: would is be useful to replace the explicit `np.float32` by `ldos_dtype` or something like a `DEFAULT_NP_DATA_DTYPE`, which would be a package-wide default?

Good question. I am not sure. I personally think that FP32 is the reasonable default for every application here, because torch internally uses it always (or our torch code, that is). So I'd argue for giving the FP64 only as a legacy / very very specific option to users to avoid confusion.

mala/descriptors/bispectrum.py
RandomDefaultUser: Fixed by 2c7cf122acf4eca51f632def689b93414de177a5

mala/descriptors/bispectrum.py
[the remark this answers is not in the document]
RandomDefaultUser: Fixed by 2c7cf122acf4eca51f632def689b93414de177a5

elcorto:
> > General question: would is be useful to replace the explicit `np.float32` by `ldos_dtype` or something like a `DEFAULT_NP_DATA_DTYPE`, which would be a package-wide default?
> 
> Good question. I am not sure. I personally think that FP32 is the reasonable default for every application here, because torch internally uses it always (or our torch code, that is). So I'd argue for giving the FP64 only as a legacy / very very specific option to users to avoid confusion.

Sure. I guess I ment this more in a coding-only fashion: Replace explicit occurences of `np.float32` with something like `common.parameters.DEFAULT_NP_DATA_DTYPE` which would be equal to `np.float32`, *not* exposing this as a setting, say in `common.parameters.ParametersData`. This would be a way to change that in one place if needed. But that's not a high prio feature. Would affect those lines:

```
examples/ex11_pass_single_feature.py:    x = torch.from_numpy(inputs_array[0, 0, 0, 3:].astype(np.float32))
mala/datahandling/data_handler.py:        # (i.e. float64 to float32 conversion) saving the data this way
mala/datahandling/data_handler.py:                                                 dtype=np.float32)
mala/datahandling/data_handler.py:                                                  dtype=np.float32)
mala/datahandling/data_handler.py:                                                   dtype=np.float32)
mala/datahandling/data_handler.py:                                                    dtype=np.float32)
mala/datahandling/data_handler.py:                                             dtype=np.float32)
mala/datahandling/data_handler.py:                                              dtype=np.float32)
mala/datahandling/data_handler.py:                    tmp = tmp.astype(np.float32)
mala/datahandling/data_handler.py:                    tmp = tmp.astype(np.float32)
mala/datahandling/data_handler.py:        numpy_array = numpy_array.astype(np.float32)
mala/datahandling/lazy_load_dataset.py:        self.input_data = self.input_data.astype(np.float32)
mala/datahandling/lazy_load_dataset.py:            self.output_data = self.output_data.astype(np.float32)
mala/datahandling/lazy_load_dataset_clustered.py:        input_data = input_data.astype(np.float32)
mala/datahandling/lazy_load_dataset_clustered.py:        self.input_data = self.input_data.astype(np.float32)
mala/datahandling/lazy_load_dataset_clustered.py:            self.output_data = self.output_data.astype(np.float32)
mala/network/predictor.py:                    snap_descriptors.astype(np.float32)
mala/network/predictor.py:                snap_descriptors.astype(np.float32)
test/scaling_test.py:# the float32 converisons.
test/scaling_test.py:            data = data.astype(np.float32)
test/scaling_test.py:            data2 = data2.astype(np.float32)
test/tensor_memory_test.py:        loaded_array_raw = loaded_array_raw.astype(np.float32)
```

elcorto (commented):
This can be made shorter by `some_dict.get(key, default_value)`. That returns `default_value` if `key not in some_dict`.

RandomDefaultUser:
> This can be made shorter by `some_dict.get(key, default_value)`. That returns `default_value` if `key not in some_dict`.

Good point, I should have done that to begin with, there were also other instances where I used these clunky if-s. Fixed with c783ab2615e6a931f24b0983f3865f72f9fc48b0.

elcorto:
In terms of saving space by using float32, #436 is maybe an even better option for the future.

Issue #171, opened 2021-07-09
Make ensemble trainings natively possible


RandomDefaultUser:
I think it is doable to make training of ensemble networks natively possible. My idea would be the following:

- Create a new parameter "parallel_execution", that by default is set to 1
- If the parameter is anything but 1, upon initialization of the Parameters() object, it will try to initialize a MPI library and get no. of ranks and rank
- Should this not succeed, serial execution is assumed
- If this succeeds, the same script will be performed nr_ranks times, in parallel without communication between the ranks
- saving of parameters, network parameters etc. will be made parallel-safe, i.e. through "_rank" in the file name
- At the end the user has to only write ONE script but can use it to train an ensemble of networks by simply requesting the ressources from slurm and doing:

`mpirun -np nr_ranks training.py`

and editing `training.py` to include something like

`params.parallel_execution=5`

RandomDefaultUser:
I would close this PR for now since I think this has not been an issue of ongoing development.

PR #441, opened 2023-04-12
Implemented basic NN ensemble class 

RandomDefaultUser:
This PR implements a simple class that allows multi-model inference. There is probably some refinement needed before we can merge it back into MALA proper, but I needed a fast way to calculate this inference for a Monte-Carlo sampling test I am conducting.


mala/network/multi_predictor.py
```diff
@@ -0,0 +1,342 @@
+"""Tester class for testing a network."""
```
elcorto, line 1: This file is largely a copy of `mala/network/predictor.py`, as such many docs/comments may or may not apply (e.g. "Tester class for testing a network." at the top, which is also in `predictor.py`).

The main diff is `load_run()` and some changes in `_forward_snap_descriptors()`. There is large code duplication which can become a maint issue later. I *don't think* a solution is to let `Predictor` support either a `network` or a list of them. A subclass is fine.

As far as I could diff things, there is *no diff* in `predict_from_qeout()` and `predict_for_atoms()` at all, so they can be removed from the subclass? Changes in `_forward_snap_descriptors()` are small, maybe one could put the common code in an extra method `_forward_snap_descriptors_core()` (bad name I know).

I know this is test code to get going quickly, therefore copy-hack-run is fine. Only once this is on `develop`, it is kind of a feature and someome has to maintain it :)

mala/interfaces/ase_calculator_uncertainty.py
```diff
@@ -0,0 +1,157 @@
...
+        - [0]: A type of additional calculation data
+        - [1]: A path to it.
+
+        With this additonal calculation data (preferably from the training of
+        the neural network), calculator can access all important data such as
+        temperature, number of electrons, etc. that might not be known simply
+        from the atomic positions.
+    """
+
+    implemented_properties = ['energy', 'forces']
```
elcorto, line 41: Forces are not supported.

mala/interfaces/ase_calculator_uncertainty.py
```diff
@@ -0,0 +1,157 @@
...
+        if self.mala_parameters.targets.target_type != "LDOS":
+            raise Exception("The MALA calculator currently only works with the"
+                            "LDOS.")
+
+        if predictor is None:
+            predictor = MultiPredictor(params, networks, data)
+        else:
+            predictor = predictor
+
+        super(MALAUncertainty, self).__init__(params, networks[0],
```
elcorto, line 56: `super().__init__(...)` does it.

mala/interfaces/ase_calculator_uncertainty.py
```diff
@@ -0,0 +1,157 @@
+"""ASE calculator for MALA predictions."""
```
elcorto, line 1: This file probably started as a copy of `mala/interfaces/ase_calculator.py`, as such many docs/comments may or may not apply, e.g. `"ASE calculator for MALA predictions."`.

mala/interfaces/ase_calculator_uncertainty.py
```diff
@@ -0,0 +1,157 @@
...
+
+from ase.calculators.calculator import Calculator, all_changes
+import numpy as np
+
+from mala import Parameters, Network, DataHandler, LDOS, MultiPredictor
+from mala.common.parallelizer import get_rank, get_comm, barrier
+from mala.interfaces.ase_calculator import MALA
+
+
+class MALAUncertainty(MALA):
```
elcorto, line 11: We plan to implement other UQ methods and I need to think about an inferface for that, but we can't steal the generic name `MALAUncertainty` here :)

I suggest `MALASimpleEnsemble`. See also the other suggested name change below: `MultiPredictor` -> `SimpleEnesmblePredictor`.

mala/interfaces/ase_calculator_uncertainty.py
```diff
@@ -0,0 +1,157 @@
...
+
+            ldos_calculator.read_from_array(ldos[m])
+            energy, self.last_energy_contributions \
+                = ldos_calculator.get_total_energy(return_energy_contributions=
+                                                   True)
+            energies.append(energy)
+            barrier()
+
+
+        # Use the LDOS determined DOS and density to get energy and forces.
```
elcorto, line 138: That comment doesn't apply.

mala/interfaces/ase_calculator_uncertainty.py
```diff
@@ -0,0 +1,157 @@
...
+        self.results["energy"] = np.mean(energies)
+        self.results["energy_uncertainty"] = np.std(energies)
+        # if "forces" in properties:
```
elcorto, line 141: It would be useful to keep `energies` for external access, e.g. `results["energy_samples"]`.

mala/interfaces/ase_calculator_uncertainty.py
```diff
@@ -0,0 +1,157 @@
...
+            energy, self.last_energy_contributions \
+                = ldos_calculator.get_total_energy(return_energy_contributions=
+                                                   True)
+            energies.append(energy)
+            barrier()
+
+
+        # Use the LDOS determined DOS and density to get energy and forces.
+        self.results["energy"] = np.mean(energies)
+        self.results["energy_uncertainty"] = np.std(energies)
```
elcorto, line 140: Uncertainty is not a well-defined quantity. Better call it `energy_std`.

mala/network/multi_predictor.py
```diff
@@ -0,0 +1,342 @@
...
+
+from mala.common.parallelizer import printout, get_rank, barrier
+from mala.network.predictor import Predictor
+from mala.network.network import Network
+from mala.datahandling.data_scaler import DataScaler
+from mala.datahandling.data_handler import DataHandler
+from mala import Parameters
+
+
+class MultiPredictor(Predictor):
```
elcorto, line 23: I'd call it `SimpleEnesmblePredictor`. I have no super strong attachment to the `SimpleEnesmble` name as such. We should just

* make sure to clarify that this is *not* the method termed Deep Ensemble from https://proceedings.neurips.cc/paper/2017/hash/9ef2ed4b7fd2c810847ffa5fa85bce38-Abstract.html / https://arxiv.org/abs/1612.01474 [a poor name btw, but that's another story], since we do not learn a noise model but just have the epistemic uncertainty part from a .. well .. simple ensemble of models
* have the same method name here and in the ASE interface

mala/network/multi_predictor.py
```diff
@@ -0,0 +1,342 @@
...
+        List of networks used for predictions.
+
+    data : mala.datahandling.data_handler.DataHandler
+        DataHandler, in this case not directly holding data, but serving
+        as an interface to Target and Descriptor objects.
+    """
+
+    def __init__(self, params, networks, data):
+        # copy the parameters into the class.
+        super(MultiPredictor, self).__init__(params, networks[0], data)
```
elcorto, line 44: `super().__init__(...)` does it.

mala/interfaces/__init__.py
```diff
@@ -1,2 +1,3 @@
 """Interfaces to other codes for workflow setup (e.g. MD or MC)."""
 from .ase_calculator import MALA
+from .ase_calculator_uncertainty import MALAUncertainty
```
elcorto, line 3: Can this class be moved to the `ase_calculator` module, or what is the reason for having a separate module?

mala/network/__init__.py
```diff
@@ -7,6 +7,7 @@
 from .hyper_opt_naswot import HyperOptNASWOT
 from .hyper_opt_oat import HyperOptOAT
 from .predictor import Predictor
+from .multi_predictor import MultiPredictor
```
elcorto, line 10: Can this class be moved to the `predictor` module, or what is the reason for having a separate module?

elcorto:
Does this close #171 or is it a step into this direction?

examples/ex19_ase_calculator_with_uncertainty.py
```diff
@@ -0,0 +1,44 @@
...
+
+import mala
+from mala import printout
+from ase.io import read
+
+from mala.datahandling.data_repo import data_repo_path
+data_path = os.path.join(data_repo_path, "Be2")
+
+"""
+ex15_ase_calculator.py: Shows how MALA can be used as an ASE calculator.
```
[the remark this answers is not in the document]
RandomDefaultUser: Good point, done.

examples/ex19_ase_calculator_with_uncertainty.py
```diff
@@ -0,0 +1,44 @@
...
+REQUIRES LAMMPS AND QUANTUM ESPRESSO (TOTAL ENERGY MODULE).
+"""
+
+
+# Load a model.
+assert os.path.exists("be_model.zip"), "Be model missing, run ex01 first."
+
+# Set up the ASE objects.
+atoms = read(os.path.join(data_path, "Be_snapshot1.out"))
+calculator = mala.MALAUncertainty.load_model("be_test_model*")
```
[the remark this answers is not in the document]
RandomDefaultUser: I have not uploaded these yet - I will once the development on this branch is finalized.

mala/network/multi_predictor.py
RandomDefaultUser: Good point, I reduced the code duplication!

mala/interfaces/ase_calculator_uncertainty.py
RandomDefaultUser: This is what my IDE gives by default - I was under the impression that doing so is preferred because it is safe even for instances where there are multiple parent classes?

mala/interfaces/ase_calculator_uncertainty.py
RandomDefaultUser: That's correct, I will have to check once development finishes that I do not miss any comments or docstrings.

mala/interfaces/ase_calculator_uncertainty.py
RandomDefaultUser: Yeah, good point. I will change the name.

mala/interfaces/ase_calculator_uncertainty.py
RandomDefaultUser: True, deleted the comment.

mala/interfaces/ase_calculator_uncertainty.py
RandomDefaultUser: Good point. I implemented this now.

mala/interfaces/ase_calculator_uncertainty.py
RandomDefaultUser: Good point. I implemented this.

mala/network/multi_predictor.py
RandomDefaultUser: Yeah, that makes sense. I like simple ensemble and renamed the class accordingly.

mala/network/multi_predictor.py
RandomDefaultUser: See above - I thought this was the correct way of doing things.

mala/interfaces/__init__.py
RandomDefaultUser: I was under the impression that one class per file is best practice?

mala/network/__init__.py
RandomDefaultUser: See above.

mala/interfaces/ase_calculator_uncertainty.py
elcorto: It is valid code. However, in the vast majority of cases, especially in simple single inheritance, `super()` is usually preferred as it will produce the same method resolution order (MRO) as the more verbose `super(DerivedClass, self)` does, but it makes code more readable and maintainable, e.g. should `DerivedClass` be renamed. But it is a matter of style.

In complicated multiple inheritance cases, one may pass arguments to `super()` in order to influence the MRO. See also https://docs.python.org/3/library/functions.html#super .

mala/network/multi_predictor.py
elcorto: Same as above, not wrong but a matter of preference.

mala/interfaces/__init__.py
elcorto: Not sure if that is a best practice. It is a design descision and here it makes sense to make the separation more visible, I guess.

mala/network/__init__.py
elcorto: Yes as above, it is one possible design decision (`module_method_a.ClassMethodA`).

examples/ex19_ase_calculator_with_uncertainty.py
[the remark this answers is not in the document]
elcorto: OK, thanks. I'll leave this unresolved as a reminder? If this is not needed, please go ahead and resolve the conversation.

elcorto:
> > Does this close #171 or is it a step into this direction?
> 
> Not directly. #171 talks about training of such ensembles, this is about inference. But maybe the same issue can be tackled in this PR.

OK, I see. Yes, we should discuss if/how much of this can be automated. ATM the "interface" this PR proposes is to have N trained models on disk, if I understand correctly.

Note that some UQ methods require a modification of the model and/or the training: 

* simple ensembles require N train runs (ideally to the same test loss)
* "deep ensembles" require additional output nodes, so model modification
* dropout requires training *with* dropout, so a modification of the model and additional hyper params (number and location of Dropout layers, dropout rate(s))
* Laplace as postprocessing needs one single train run, no modification of model and train setup
* etc ...

So there are some design things to figure out. Probably we'll need to modify the UQ-related API a bit as we go forward and implement multiple UQ methods.

Given the above, we may consider not actually merging to `develop` (where we kind of fix the API and names that we later need to support) but to another branch, e.g. `feature-uncertainty`.

PR #442, opened 2023-04-21
Generalize Tester class to test training and validation partitions

kyledmiller:
Added an optional argument to predict_targets() and test_snapshot() allowing the user to specify 'tr' and 'va' partitions for testing.

This allows the user to easily calculate unscaled training and validation predictions for comparing accuracy between data set partitions.


RandomDefaultUser:
Looks good to me, merging it!

PR #449, opened 2023-05-23
CI for openPMD

franzpoeschel:
This adds openPMD-api to the CI infrastructure and adds a test for openPMD. At the moment, that test is `test/shuffling_openpmd_test.py` which tests openPMD-based shuffling.

It would be wise to add also a test that covers the basic functionality of openPMD I/O, especially also in parallel. Is there a test that uses numpy so far which I can modify to that purpose?


test/shuffling_openpmd_test.py
```diff
@@ -0,0 +1,179 @@
...
+                                               data_handler.output_dimension]
+        test_network = mala.Network(test_parameters)
+        test_trainer = mala.Trainer(test_parameters, test_network,
+                                    data_handler)
+        test_trainer.train_network()
+        old_loss = test_trainer.final_validation_loss
+
+        # Shuffle.
+        test_parameters.data.shuffling_seed = 1234
+        data_shuffler = mala.DataShuffler(test_parameters)
```
franzpoeschel, line 146: I think that reusing the `test_parameters` in such a way might be dangerous, as the old `parameters.snapshot_directories_list` will be taken over. This issue already exists in `shuffling_test.py` based on numpy.

RandomDefaultUser: Good point, I can change that. That should be an easy fix.

test/shuffling_test.py
```diff
@@ -48,6 +49,64 @@ def test_seed(self):
...
+        test_parameters = mala.Parameters()
+        test_parameters.openpmd_configuration = {
+            'adios2': {
+                'engine': {
+                    'parameters': {
+                        'StatsLevel': 1
+                    }
+                }
+            }
+        }
```
franzpoeschel, line 63: Given that we are using HDF5, we can remove the ADIOS2 config here and everywhere else. I'll push a commit.

test/shuffling_test.py
```diff
@@ -48,6 +49,64 @@ def test_seed(self):
         new = np.load("Be_REshuffled1.out.npy")
         assert np.isclose(np.sum(np.abs(old-new)), 0.0, atol=accuracy)
 
+    def test_seed_openpmd(self):
+        """Test that the shuffling is handled correctly internally."""
```
franzpoeschel, line 53: If I understand this correctly, this test makes sure that Numpy- and openPMD-shuffling give the same results?
In that case, I would maybe make the documentation a bit more concrete here.

test/shuffling_openpmd_test.py
```diff
@@ -1,179 +0,0 @@
...
-        data_handler.prepare_data()
-
-        test_network = mala.Network(test_parameters)
-        test_trainer = mala.Trainer(test_parameters, test_network,
-                                    data_handler)
-        test_trainer.train_network()
-        new_loss = test_trainer.final_validation_loss
-        assert old_loss > new_loss
-
-TestDataShuffling().test_training()
```
franzpoeschel: Just for being sure: This test is now integrated into the workflow and the normal shuffling test?

franzpoeschel (commented):
Looks good to me, I have some inline comments.
Given that parallel shuffling is now merged, should we maybe test that, too?

test/shuffling_openpmd_test.py
RandomDefaultUser: Yes!

test/shuffling_test.py
RandomDefaultUser: Makes sense. I fixed that.

DanielKotik:
@franzpoeschel @RandomDefaultUser If openPMD is an optional dependency for MALA, it should not be part of `requirements.txt` (instead add it [here](https://github.com/mala-project/mala/blob/8b328769201b197d3cb35dcf692e3d219c999df2/setup.py#L19)) and should only be added explicitly in the `Dockerfile`. If it is not an optional dependency, it should be present in `requirements.txt` *and* both `mala_cpu_environment.yml` and `mala_cpu_base_environment.yml` must then be updated accordingly. It must also be removed from the `Dockerfile`.

franzpoeschel:
fixed by https://github.com/mala-project/mala/pull/541

PR #450, opened 2023-05-23
[WIP] Enabling GPU calculation of descriptors via LAMMPS

### LINKED ISSUE #399: Port LAMMPS descriptor calculation to GPU
For more efficient inference, it would be advantageous to have a GPU accelerated version of the descriptor version of LAMMPS. 
This includes these steps:

1. Check that the version of LAMMPS we use (https://github.com/athomps/lammps/tree/mala) is in principle working on GPU
2. Test that the descriptor calculation still gives correct results when compiled with GPU support
3. Move descriptor calculation (i.e., all `compute_grid_local` calls) to GPU, if they are not there by default (which they shouldn't be, at least not fully)

RandomDefaultUser:
This PR will close #399. Based on https://github.com/mala-project/lammps/pull/1, which implements the LAMMPS side of things, this PR essentially just tells MALA to use the correct LAMMPS parameters and input scripts to use the GPU (via Kokkos) when calculating descriptors. 

To-Dos:

- [ ] Get the Kokkos-LAMMPS version to work with MALA - currently the separately build executable works, but not the library (the `-k` flag is not recognized, plus I get a segfault)
- [ ] Document LAMMPS build with Kokkos+GPU
- [ ] Test results

### DIFF
--- a/mala/descriptors/atomic_density.py
+++ b/mala/descriptors/atomic_density.py
@@ -140,12 +140,8 @@ class AtomicDensity(Descriptor):
             self.parameters.atomic_density_sigma = self.\
                 get_optimal_sigma(voxel)
 
-        # Build LAMMPS arguments from the data we read.
-        lmp_cmdargs = ["-screen", "none", "-log",
-                       os.path.join(outdir, "lammps_ggrid_log.tmp")]
-
         # LAMMPS processor grid filled by parent class.
-        lammps_dict = self._setup_lammps_processors(nx, ny, nz)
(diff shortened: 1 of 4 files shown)


RandomDefaultUser:
This PR is  finished. What is currently missing is documentation on how to use the LAMMPS GPU interface in the code, but I am waiting until the total energy module + GPU has been finished before writing this up, since the docs need to be overhauled in general in this regard and I'd like to have the full picture.

PR #458, opened 2023-05-26
openPMD: shuffle data in parallel

franzpoeschel:
The data shuffling is relatively trivial to parallelize over the newly created shuffled checkpoints.
Processes cannot easily cooperate in the creation of one single checkpoint as the shuffling operation is global within one checkpoint.
So, this PR distributes the to-be-created checkpoints over parallel workers. The input checkpoints are accessed in parallel. This parallel access does not really scale to full system size, but since there is only ever a mediocre number of checkpoints, such scaling is not really necessary.
An alternative would be to access input checkpoints one after another in parallel, but this has performance impacts (checkpoints might be opened and closed multiple times if multiple shuffled checkpoints are created per process), and processes would have to ensure that each process opens the checkpoints exactly the same amount of time.
---> Easier to just open everything, since it's fine to do so anyway.

### DIFF
--- a/mala/common/physical_data.py
+++ b/mala/common/physical_data.py
@@ -338,6 +338,7 @@ class PhysicalData(ABC):
         ----------
         path : string
             File to save into. If no file ending is given, .h5 is assumed.
+            Alternatively: A Series, opened already.
 
         array : Either numpy.ndarray or an SkipArrayWriting object
             Either the array to save or the meta information needed to create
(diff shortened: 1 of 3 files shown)


mala/datahandling/data_handler_base.py
```diff
@@ -148,7 +148,7 @@ def clear_data(self):
     # Loading data
     ######################
 
-    def _check_snapshots(self):
+    def _check_snapshots(self, comm=None):
```
franzpoeschel, line 151: Using a parameter here on purpose and not `get_comm()`, as this makes it obvious at the call site that the function is collective.

mala/common/physical_data.py
```diff
@@ -242,7 +242,7 @@ def read_dimensions_and_dtype_from_numpy_file(self, path):
         loaded_array = np.load(path, mmap_mode="r")
         return self._process_loaded_dimensions(np.shape(loaded_array)), loaded_array.dtype
 
-    def read_dimensions_from_openpmd_file(self, path):
+    def read_dimensions_from_openpmd_file(self, path, comm=None):
```
franzpoeschel, line 245: Using a parameter here on purpose and not `get_comm()`, as this makes it obvious at the call site that the function is collective.

mala/datahandling/data_shuffler.py
```diff
@@ -277,6 +321,9 @@ def from_chunk_i(i, n, dset, slice_dimension=0):
                     shuffle_dimensions)
             shuffled_snapshot_series.close()
 
+        for series in input_series_list:
```
franzpoeschel, line 324:
```suggestion
        # Ensure consistent parallel destruction
        # Closing a series is a collective operation
        for series in input_series_list:
```

RandomDefaultUser:
Looks good to me, I will be merging this now!

PR #462, opened 2023-06-17
Implement Atomic forces

RandomDefaultUser:
- [x] Implement and test dE[D]/dd 
- [ ] Implement dE[n]/dn (may require some changes in QE)
- [x] Implement dD/dd (this should be trivial)
- [x] Implement  dn/dd (this should be trivial)
- [x] Implement nd/nB 

   - Essentially working, as this is trivial with torch; but there is some changes to the internal workflow required here
- [x] Implement nB/nR (in LAMMPS)


RandomDefaultUser:
Closed in favor of #679 so that the comparing branch no longer lives in my fork.

Issue #454, opened 2023-05-23
Installation breaks when using opt setup flag (pqkmeans, oapackage cannot be installed)


karanprime:
I wanted to test how clustering and sampling according to cluster sizes affects the training. However, the package required for that, pqkmeans, cannot be installed from pip, nor by building it from scratch.

Reproduction:
In a fresh environment, I tried installing mala with the optional flag `pip install -e .[opt]`. 

This results in an error `RuntimeError: numpy cannot be imported. numpy must be installed prior to installing OApackage`

For the clustering part, I only need pqkmeans. So I tried to install it separately using `pip install pqkmeans`.
This results in an error for a dependency of pqkmeans called lshash3.
`ERROR: Failed building wheel for lshash3`.
lshash3 hasn't been updated on pypi since 2017.

I then tried building pqkmeans using instructions from their [github page](https://github.com/DwangoMediaVillage/pqkmeans) but that also throws errors. 

OApackage pypi page: https://pypi.org/project/OApackage/
pkqmeans pypi page: https://pypi.org/project/pqkmeans/
lshash3 pypi page: https://pypi.org/project/lshash3/

tl:dr; Installation for OApackage and pqkmeans is broken.

karanprime:
After the hackathon, we discussed the following todos:
- Fix OAPackage dependency
- Remove clustering code from the main code
I will open a PR for this.

PR #464, opened 2023-07-04
Fix `pkqmeans` and `oapackage` install

RandomDefaultUser:
As discussed in issue #454, currently `oapackage` and `pkqmeans`are not working and should be adapted/removed from the install script. 

To-Dos:

- [x] Remove clustering from the code base altogether, because it is not needed at the moment; this makes the installation of `pkqmeans` irrelevant
- [x] Fix the installation of `oapackage`


RandomDefaultUser:
I just checked the installation of `oapackage` and it works now. Judging from the PyPI page there were some recent updates (the project is still under active development): https://pypi.org/project/OApackage/#history and I guess one of those fixed it.

PR #471, opened 2023-07-30
MALA v1.2.0

RandomDefaultUser:
With this release, we'd like to get MALA more production ready: better accessibility for users, better performance during both training and inference, etc.


acangi (approved):

srajama1 (approved):
Thanks @RandomDefaultUser !

PR #472, opened 2023-08-01
Fix possible TEM build silent fail

### LINKED ISSUE #469: `build_total_energy_module.sh` can silently fail
The `grep` in  

```sh
pwobjs="$(find $pw_src_path -name "*.o" | grep -E -v 'generate_vdW_kernel_table|generate_rVV10_kernel_table')"
```

will exit 1 when there is no match. When using `set -euo pipefail` we silently exit here via `set -e`. The most simple fix would be to use just `set -u`. This will make the script run through even if commands fail in between, but I guess that's ok for this small piece of code.

elcorto:
Closes #469

### DIFF
--- a/external_modules/total_energy_module/build_total_energy_module.sh
+++ b/external_modules/total_energy_module/build_total_energy_module.sh
@@ -15,10 +15,7 @@ err(){
 
 [ $# -eq 1 ] || err "Please provide exactly one argument (the path to the QE directory)" && root_dir=$1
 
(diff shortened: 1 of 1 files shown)


elcorto:
CPU tests / cpu-tests failed b/c of https://github.com/mala-project/test-data/issues/6. However building locally works.

RandomDefaultUser (approved):
Looks good to me, this can be merged.

PR #493, opened 2023-11-27
Hotfixing a problem with DOS comparison in the Tester class

RandomDefaultUser:
@timcallow found the bug, this should fix it.

### DIFF
--- a/mala/network/tester.py
+++ b/mala/network/tester.py
@@ -293,10 +293,16 @@ class Tester(Runner):
 
             target_calculator.read_from_array(predicted_target)
             predicted = target_calculator.density_of_states
-            return np.mean(np.abs((actual - predicted) / actual)) * 100
 
+            percentage_error = 0
+            values_counted = 0
+            for i in range(0, self.parameters_full.targets.ldos_gridsize):
+                if actual[i] > 0:
+                    percentage_error += np.abs((actual[i] - predicted[i]) / actual[i]) \
+                                        * 100
+                    values_counted += 1
 
-
+            return np.ma.masked_invalid(np.abs((actual - predicted) / actual)).mean() * 100
 
     def __prepare_to_test(self, snapshot_number):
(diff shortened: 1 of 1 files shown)


timcallow:
@RandomDefaultUser sorry for the late response. It seems the function is working as intended, i.e. there is no bug. In that respect it is ready to be merged.

However, there is more a conceptual issue with the choice of error function. I noticed this when training models with mixed numbers of bands, in this 576 and 600 (this wasn't really deliberate, just sort of happened). In this case, the DOS is always basically "very good" but the energy cut-off for the DOS differs. This leads to a sort of asymmetry in the error depending on whether it's tested on a snapshot with 576 or 600 bands, which is best seen from the attached images. 

One of the errors seen in these images is also quite large - 20%. This is entirely caused by the different cut-offs. It also doesn't affect the physics of the problem: consider the images where the DOS is multiplied by the occupations. As the Fermi energy is lower than the DOS cut-off energy, these different cut-offs don't affect the DOS * occupations, which is actually of more importance physically.

In any case, I'd say it can be merged, since technically it does work as expected. However we might want to think about defining it differently in future, and maybe it's worth raising an issue (feel free to assign it to me).

Screenshot from 2023-12-13 17-34-43
Screenshot from 2023-12-13 17-34-25
Screenshot from 2023-12-13 17-35-12
Screenshot from 2023-12-13 17-34-58

RandomDefaultUser:
Thanks for the feedback @timcallow ! Should we maybe revert to the symmetric MAPE again? We already had that in the middle of the bugfixing process, maybe it is wise to go back? That would fix the problem, wouldn't it?

RandomDefaultUser:
After some additional discussion with @timcallow we decided to implement a symmetric MAPE with a small number added to both predicted and actual value of the DOS to reduce numerical noise. 
While not optimal, I think this quantity is the best option we have for DOS comparisons at the moment.

PR #494, opened 2023-12-08
Update CITATION.cff

acangi:
As discussed, I'm changing the order of the author list in the citation file. Putting the two scientific leads in front (me and Siva). The remaining list is then in alphabetical order. 
We could, however, decide on a different order after the scientific leads. Let me know what you think.


RandomDefaultUser:
Thank you for this update, this is a really good idea, since I think it makes the most sense to have permanent staff members and scientific leads at the front, despite the overall alphabetic ordering. LGTM, I will merge this!

PR #499, opened 2024-01-31
v1.2.1 - Minor Bugfixes

RandomDefaultUser:
This PR fixes some minor issues and bugs, and updates some of the meta information. It also serves as a point of reference for an upcoming scientific work. 

# Change notes:

- Updated MALA logos
- Updated Tester class to also give Kohn-Sham energies alongside LDOS calculations 
- Updated CITATION.cff file to reflect new team members and scientific supervisors
- Fixed bug that would crash models trained with horovod when loaded for inference without horovod
- Fixed bug that would crash training when using Optuna+MPI for hyperparameter optimization (GPU compute graph usage had not been properly adapte for this scenario)
- Deactivated pytorch profiling by default, can still be manually enabled


srajama1 (approved):
Thanks Lenz! @RandomDefaultUser

acangi (approved):

Issue #502, opened 2024-02-13
SNAP data overwritten for simultaneous pre-processing in same directory


timcallow:
When pre-processing SNAP data with lammps, there is an issue if multiple jobs are launched simultaneously from the same directory. The data for one snapshot is sometimes overwritten by the data corresponding to a different snapshot, because of the way mala calls lammps. 

In the best case, this behaviour should be fixed so that multiple pre-processing jobs can be invoked simultaneously in the same directory. In the meantime, a warning to the user should be printed.

RandomDefaultUser:
Closed with #518

PR #503, opened 2024-02-13
Add warning for lammps pre-processing

timcallow:
Warning for users when pre-processing SNAP with lammps.  This is a temporary workaround for issue #502

### DIFF
--- a/mala/descriptors/descriptor.py
+++ b/mala/descriptors/descriptor.py
@@ -526,7 +526,7 @@ class Descriptor(PhysicalData):
         """
         from lammps import lammps
 
-        printout("Warning: do not initialize more than one pre-processing calculation\
+        parallel_warn("Do not initialize more than one pre-processing calculation\
 in the same directory at the same time. Data may be over-written.")
 
         if self.parameters._configuration["mpi"] and \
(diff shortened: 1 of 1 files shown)


mala/descriptors/descriptor.py
```diff
@@ -526,6 +526,9 @@ def _setup_lammps(self, nx, ny, nz, outdir, lammps_dict,
         """
         from lammps import lammps
 
+        printout("Warning: do not initialize more than one pre-processing calculation\
+in the same directory at the same time. Data may be over-written.")
```
RandomDefaultUser, line 530:
```suggestion
parallel_warn("Do not initialize more than one pre-processing calculation\
in the same directory at the same time. Data may be over-written.")
```

RandomDefaultUser (changes requested):
Thanks for the PR @timcallow, I agree that for now (until we have fixed the underlying issue) a warning like this is the best way to proceed. MALA has a parallel warning function that I would suggest we use instead of the printout function.

timcallow:
Thanks for the suggestion, I didn't notice that function - I've committed your change

RandomDefaultUser (approved):
Thanks, looks good to me now! :)

PR #517, opened 2024-05-30
Resolving pytest issues

pcagas:


pcagas:
- Calling the tests with `runpy` now uses a path relative to the test file and not the current directory.
- The tests are run from `pytest` temp directory so the artifacts don't pollute the current one.

pcagas:
This should _not_ be merged before https://github.com/mala-project/test-data/pull/9

RandomDefaultUser (changes requested):
Thanks for the PR, @pcagas ! That in principle looks all great to me, and I think it is almost ready to merge. I would, however, move the newly added model in the data repository (as mentioned in https://github.com/mala-project/test-data/pull/9#pullrequestreview-2099651701), which then in turn requires an update to the reference model file in some of the tests. 

The ones that need updating and fail elsewise are:
 
Screenshot from 2024-06-05 17-58-03


Also, the data repository has moved to a RODARE mirror system now. 

So my suggestion would be to 

1. Adjusting https://github.com/mala-project/test-data/pull/9 as per my review
2. Merging https://github.com/mala-project/test-data/pull/9
3. Creating a new release for https://github.com/mala-project/test-data/ 
4. Merge the MALA develop branch into this branch, which should give you the code for downloading from RODARE
5. Adjust the download section in the workflow file to point to the correct version:
updating_data_repository
6. Fix the tests that break due to the updated model path (see above)

Then we should be good to merge!

pcagas:
- [ ] Adjusting https://github.com/mala-project/test-data/pull/9 as per @RandomDefaultUser review
- [ ] Merging https://github.com/mala-project/test-data/pull/9
- [ ] Creating a new release for https://github.com/mala-project/test-data/
- [ ] Merge the MALA develop branch into this branch, which should give you the code for downloading from RODARE
- [ ] Adjust the download section in the workflow file to point to the correct version
- [ ] Fix the tests that break due to the updated model path

RandomDefaultUser (approved):
Looks good to me, can be merged!

PR #520, opened 2024-05-30
Fix requeue zombie trials

### LINKED ISSUE #497: Optuna v3.x.x no longer compatible with zombie trial cleaning
MALA implements a way to clean "zombie trials", i.e. trials which are not expected to finish anymore (e.g., because the original job broke). The API for this has changed from optuna v3 to v2, and this function no longer works. The regular MALA-optuna interface still works, but this particular part is currently not-functional. As it is important for extended, distributed hyperparameter optimization, I marked this as "critical". 
Fixing this should not be too hard.

elcorto:
Closes #497 

The test works, the only strange thing which is a bit unclear is documented in [this comment](https://github.com/mala-project/mala/blob/acff38534c8a0c4c7fcf1ee7a469f26b99343b5d/test/hyperopt_test.py#L464-L482).

### DIFF
--- a/mala/network/hyper_opt_optuna.py
+++ b/mala/network/hyper_opt_optuna.py
@@ -176,9 +176,18 @@ class HyperOptOptuna(HyperOpt):
         cleaned_trials = []
         for trial in trials:
             if trial.state == optuna.trial.TrialState.RUNNING:
-                study_to_clean._storage.set_trial_state(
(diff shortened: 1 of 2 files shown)


elcorto:
Comment updated and resolved.

RandomDefaultUser (approved):
Looks good to me, thank you!

PR #524, opened 2024-05-30
Update docs workflow and dependencies

DanielKotik:
I was not able to create the documentation locally with a recent Python 3.11 installation. The versions of some sub-dependencies differ between Python 3.8 and Python 3.11. One of the sub-dependencies from packages specified in `docs/requirements.txt` expects a more recent Sphinx version `>=5`:

```sh
Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
```

This PR fixes this and other (`Node.js`) issues:

image

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -14,10 +14,10 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Check out repository
-        uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


RandomDefaultUser (approved):
This implicitly sets a python version requirement for python if the documentation is locally built. That is fine for now, but we need to be aware of it and eventually properly set a required python version for MALA.

PR #530, opened 2024-05-31
DataConverter class: Add reading logic for openPMD and numpy

franzpoeschel:
This is supposed to make the transition from Numpy to openPMD more smooth as the `DataConverter` class can now be used to convert between Numpy and openPMD ad libitum.

TODO:

- [x] This logic only applies to the bare array. Since converting between Numpy and openPMD means that either the data source or sink has no metadata, there is no specific logic to write metadata, but instead only the bare array is taken care of. Further metadata must be defined by the user in the conversion script. **Should we do more than that?**
- [x] The example for now uses some local data that I have on my disk, use the sample data instead.
- [ ] Maybe provide a bare CLI interface.


RandomDefaultUser:
Thanks for all the great work @franzpoeschel , I will be merging this now!

Issue #165, opened 2021-07-02
Future data repositories - where to store them?


DanielKotik:
We will have large data repositories in the future and we need to discuss where to host them.

Background information:

We have a small (~ 33MB) [test data repository](https://github.com/mala-project/test-data) with binary data (Numpy arrays, pickled Python objects etc.) For good reasons such binary data should usually not reside inside a git repository. If this is unavoidable or useful for other reasons one should then make use of [git lfs](https://git-lfs.github.com/). At first we have done so for our test data repo. GitHub however not only puts limits on the lfs storage quota (which is of course higher than for a normal git repo):

![grafik](https://user-images.githubusercontent.com/28047702/124280917-824b4a80-db49-11eb-8794-0b1b3db91a65.png)

but also on the bandwidth (1 GB of bandwidth per month). This has already prevented us from the use of our test data repo within the cpu test pipeline. For this reason and because of the small size of our binary test data, we've moved again away from using git lfs.

For future data repositories with huge amount of data (~GBs) we have to think about alternative hosting services. Things to keep in mind are:

- documentation of the repositories content (that's why it's a bad idea to store data inside a docker container which lives somewhere in a registry - the data is almost hidden, where to put a description of the data?)
- reproducible research, citability (getting a DOI from Zenodo, RODARE, figshare or some other more data specialised hosting service)
- version control (git lfs, [DVC](https://dvc.org/))
- FAIR data: Findability, Accessibility, Interoperability and Reusability

There is a lot of literature out on this topic; to list a few:

- https://web.archive.org/web/20190503003656if_/https://towardsdatascience.com/why-git-and-git-lfs-is-not-enough-to-solve-the-machine-learning-reproducibility-crisis-f733b49e96e8?gi=ec1787bd9ed6
- https://www.nature.com/sdata/policies/repositories#physics

RandomDefaultUser:
I think we need to prioritize this higher up; In its current form, the test data repo may have a LDOS, but it is by nature inaccurate due to the size limitations. If we had a bigger data repo, we could host physically correct data.

acangi:
Our default data repository is [RODARE](https://rodare.hzdr.de). While it is well integrated with our systems, it has the drawback that an individual file cannot be accessed, but the entire dataset needs to be downloaded. 

An alternative to RODARE might be the Jülich repository [b2share](https://b2share.fz-juelich.de/). It is similar to RODARE and allows for accessing individual files of a larger dataset. 

[...]

PR #534, opened 2024-06-02
Added RODARE download

RandomDefaultUser:
This PR enables download of data repository from RODARE instead of Github, to avoid download quoat issues.

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -171,13 +171,20 @@ jobs:
           # `requirements.txt` and/or extra dependencies are missing in the Docker Conda environment
           diff env_1.yml env_2.yml
 
-      - name: Check out repository (data)
-        uses: actions/checkout@v3
-        with:
-          repository: mala-project/test-data
-          path: mala_data
-          ref: v1.7.0
-          lfs: true
+      - name: Download test data repository
+        shell: 'bash -c "docker exec -i mala-cpu bash < {0}"'
+        run: |
+          # Download test data repository from RODARE. If the version changes
+          # this URL has to be adapted (the 29xx number and the version have 
+          # to be incremented)
+          wget "https://rodare.hzdr.de/record/2901/files/mala-project/test-data-v1.7.4.zip"
+          
+          # Once downloaded, we have to unzip the file. The name of the root
+          # folder in the zip file has to be updated for data repository
+          # updates as well - the string at the end is the hash of the data 
+          # repository commit. 
+          unzip test-data-v1.7.4.zip
+          mv mala-project-test-data-a6458c5 mala_data
 
       - name: Test mala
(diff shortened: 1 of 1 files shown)


DanielKotik:
@RandomDefaultUser Is there a related recent issue? For the sake of completeness, I mention there this old issue:

- #165

RandomDefaultUser:
Hi @DanielKotik there is not really an issue for precisely this. But I think the issue you mentioned can be closed with this PR.

Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apt-get --allow-releaseinfo-change update && apt-get upgrade -y && \
     build-essential \
     libz-dev \
     swig \
-    git-lfs \
+    unzip \
```
DanielKotik, line 10: I would suggest to explicitly add `wget` here as well, although it might come with the base image already.

```suggestion
    unzip \
    wget \
```

.github/workflows/cpu-tests.yml
```diff
@@ -171,13 +171,20 @@ jobs:
...
+          # Download test data repository from RODARE. If the version changes
+          # this URL has to be adapted (the number after /record/ and the 
+          # version have to be incremented)
+          wget "https://rodare.hzdr.de/record/2997/files/mala-project/test-data-v1.7.8.zip"
+          
+          # Once downloaded, we have to unzip the file. The name of the root
+          # folder in the zip file has to be updated for data repository
+          # updates as well - the string at the end is the hash of the data 
+          # repository commit. 
+          unzip test-data-v1.7.8.zip
```
DanielKotik, line 186: Add `-q` flag for quiet output to avoid cluttering the log files with noise:

```suggestion
          unzip -q test-data-v1.7.8.zip
```

.github/workflows/cpu-tests.yml
```diff
@@ -171,13 +171,20 @@ jobs:
...
+          mv mala-project-test-data-46a6992   mala_data
```
DanielKotik, line 187: Every time we update the database, we have to change something in multiple places, two places in the RODARE URL, then the version of the zip file and adapt the `mv` command as well. This is not very convenient, we should try to improve on that in the future. For now, I would leave it as is.

DanielKotik (changes requested):

Dockerfile
RandomDefaultUser: Sounds good, let's do that!

.github/workflows/cpu-tests.yml
RandomDefaultUser: Ah good point, let's do that!

.github/workflows/cpu-tests.yml
RandomDefaultUser: Good point, and while not urgent right now, this should be done in the future. Thanks for documenting this in https://github.com/mala-project/mala/issues/537.

PR #539, opened 2024-06-06
Allow CI runs for normal PRs, but avoid runs for draft PRs

### LINKED ISSUE #514: Avoid CI runs when PR is still in "draft" mode
A Draft PR should not activate the CI, this would reduce overall Github action time required.

DanielKotik:
Finally closes #514

Specifying `if: ${{ ! github.event.pull_request.draft }}` is only required for the first job in `cpu-tests.yml` and `gh-paages.yml` as subsequent jobs in the respective workflow files depend on the first job. This means that if the first job is skipped due to a PR draft, all dependent jobs are also skipped, which is what we intend.

image

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -5,6 +5,9 @@ on:
     # Trigger on pull requests to master or develop that are
     # marked as "ready for review" (non-draft PRs)
     types:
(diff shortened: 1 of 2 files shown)


RandomDefaultUser (approved):
Looks good to me, thank you!

PR #541, opened 2024-06-06
Fix CI installation of openPMD-api

### LINKED ISSUE #536: Make OpenPMD consistent
As pointed out by @DanielKotik here https://github.com/mala-project/mala/pull/449#issuecomment-2149692885 OpenPMD is currently added to the installation process/CI somewhat inconsistently.

### LINKED ISSUE #537: Improve CI with regards to data repo updates
Every time we update the data repo, we have to change something in multiple places, two places in the RODARE URL, then the version of the zip file and on top adapt the `mv` command as well. This is not very convenient, we should try to improve on that in the future. For now, I would leave it as is.

franzpoeschel:
Close #536

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,6 @@ RUN conda env create -f mala_${DEVICE}_environment.yml && rm -rf /opt/conda/pkgs
 RUN /opt/conda/envs/mala-${DEVICE}/bin/pip install --no-input --no-cache-dir \
     pytest \
     oapackage==2.6.8 \
-    openpmd-api==0.15.1 \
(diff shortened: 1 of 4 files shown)


DanielKotik:
@franzpoeschel and I have noticed that when installing the mala package in the CI, the openPMD package pre-installed via Conda is ignored and instead an installation via pip is forced, https://github.com/mala-project/mala/actions/runs/9483350602/job/26130690213#step:7:29. Therefore, we have decided to temporarily specify openPMD explicitly in the pip section of the YAML files of the Conda environment.

PR #542, opened 2024-06-10
Use RODARE api instead of hard coded URL:

### LINKED ISSUE #537: Improve CI with regards to data repo updates

DanielKotik:
It is better to use the DOI, which always refers to the latest version of the test data repo. This avoids updating the CI at several places each time there is a new version of the test data repo.

Closes  #537

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -179,7 +179,7 @@ jobs:
       - name: Download test data repository from RODARE
         shell: 'bash -c "docker exec -i mala-cpu python < {0}"'
         run: |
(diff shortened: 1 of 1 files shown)


RandomDefaultUser (approved):
Thanks a lot for this PR @DanielKotik that looks great! I am glad we can use RODARE without having to manually update the file paths all the time. 
Can be merged from my side.

PR #560, opened 2024-07-17
Unified error calculation

nerkulec:
This PR includes logging overhaul and makes energy error calculations uniform (in units of meV/atom).

### DIFF
--- a/mala/network/runner.py
+++ b/mala/network/runner.py
@@ -236,7 +236,6 @@ class Runner:
         target_calculator.read_additional_calculation_data(output_file)
 
         errors = {}
-        fe_dft = target_calculator.fermi_energy_dft
         fe_actual = None
         fe_predicted = None
         try:
--- a/mala/network/runner.py
+++ b/mala/network/runner.py
@@ -3,6 +3,8 @@
 import os
 from zipfile import ZipFile, ZIP_STORED
 
+from mala.common.parallelizer import printout
+
 import numpy as np
 import torch
 import torch.distributed as dist
@@ -10,10 +12,16 @@ import torch.distributed as dist
 import mala
 from mala.common.parallelizer import get_rank
 from mala.common.parameters import ParametersRunning
+from mala.datahandling.fast_tensor_dataset import FastTensorDataset
 from mala.network.network import Network
 from mala.datahandling.data_scaler import DataScaler
 from mala.datahandling.data_handler import DataHandler
 from mala import Parameters
+from mala.targets.ldos import LDOS
+from mala.targets.dos import DOS
+from mala.targets.density import Density
+
+from tqdm.auto import tqdm, trange
 
 
 class Runner:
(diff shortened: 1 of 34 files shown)


mala/network/runner.py
```diff
@@ -41,6 +49,334 @@ def __init__(self, params, network, data, runner_dict=None):
...
+
+        snapshot_number : int
+            Snapshot number for which the errors are calculated.
+
+        Returns
+        -------
+        errors : dict
+            Dictionary containing the errors.
+        """
+        energy_metrics = [metric for metric in metrics if "energy" in metric]
```
RandomDefaultUser, line 77: There is one metric which is not reflected in this new workflow, but which I think maybe should, that's the "_energy_full" metrics I implemented for band and total energy. The idea there was to give users the option to return either both the actual band/total energy using the Tester class. For some visualizations (e.g. the correlation plot, Fig. 4 of https://www.nature.com/articles/s41524-023-01070-z.pdf) it is nice to have the energies of all test snapshots (both DFT, actual and predicted) as one big table. 

Since the changes here unify Runner, Trainer and Tester, we would probably have to safeguard this option to only be accessible from the Tester class.

mala/network/trainer.py
```diff
@@ -912,327 +1010,6 @@ def __process_mini_batch(self, network, input_data, target_data):
...
-                                    f"batch {batchid + 1}, "  # /{total_samples}, "
-                                    f"validation avg time: {avg_sample_time} "
-                                    f"validation avg throughput: {avg_sample_tput}",
-                                    min_verbosity=2,
-                                )
-                                tsample = time.time()
-                            batchid += 1
-                    torch.cuda.synchronize(
-                        self.parameters._configuration["device"]
-                    )
```
RandomDefaultUser, line 1070: I may be missing something, but I cannot find this portion of the code in the new __validate_network - is there a reason why we would want to get rid of the validation graphs? I thought they were working nicely.

RandomDefaultUser (changes requested):
Hi @nerkulec , thank you the PR! All around, it looks pretty great, and I think unifying the error calculations is a great idea. 
I have some questions to the code (namely one maybe obscure error metric I was using occasionally and validation grpahs) that I left as comments. 

Two general things: 
1. Since this introduces the tqdm package, shouldn't we include it in the requirements.txt? Or is it reliably shipped with another package? This may also affect the cpu_environments.yml. 
2. The docs are currently failing with this PR, which is due to tqdm not yet included in the autodoc_mock_imports list in the conf.py of sphinx. Sphinx is trying to import tqdm to generate an automated API (which is of course unnecessary) but failing, since it is not installed alongside the other docs packages. By adding it to the autodoc_mock_imports list, sphinx will not attempt to import it.

mala/network/runner.py
```diff
@@ -41,6 +49,335 @@ def __init__(self, params, network, data, runner_dict=None):
...
+                    error = np.mean((predicted_outputs - actual_outputs) ** 2)
+                    errors[metric] = error
+
+                elif metric == "density":
+                    target_calculator = self.data.target_calculator
+                    if not isinstance(
+                        target_calculator, LDOS
+                    ) and not isinstance(target_calculator, Density):
+                        raise Exception(
+                            "Cannot calculate the total energy from this "
```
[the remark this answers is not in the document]
nerkulec: This is fixed now

mala/network/runner.py
```diff
@@ -41,6 +49,334 @@ def __init__(self, params, network, data, runner_dict=None):
...
+                    be_predicted = target_calculator.get_band_energy(
+                        fermi_energy=fe_predicted
+                    )
+                    be_error = (be_predicted - be_actual) * (
+                        1000 / len(target_calculator.atoms)
+                    )
+                    errors[energy_type] = be_error
+                except ValueError:
+                    errors[energy_type] = float("inf")
+            elif energy_type == "band_energy_dft_fe":
```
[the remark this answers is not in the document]
nerkulec: You're right, I removed it

mala/network/runner.py
nerkulec: I am not a big fan of functions which change their return type (here: number vs list with two numbers) based on the arguments. I'll add a separate function which provides this functionality.

mala/network/trainer.py
nerkulec: The computation graphs in validate_network only feed the inputs forward through the network, and later invoke calculate_loss on the model which is just the mean square error. No gradients are accumulated or weights updated. I could potentially leave the computation graphs here just for the LDOS metric, but then all other metrics' calculations would still run outside the computation graph, since it's outside of torch. I could re-use the network predictions out of the graph, but I doubt there is any significant speed improvement compared to evaluation in eager mode. I even doubt whether there is any noticeable improvement when you add the MSE on top. That's why I removed them altogether here. If you strongly feel that this hurts performance when using just the LDOS validation metric or done a benchmark that shows the performance difference I can put it back here.

mala/network/runner.py
nerkulec: It's now implemented in Tester.get_energy_targets_and_predictions()
Let me know if you think it should be done differently.

RandomDefaultUser: Good point, I agree that having a consistent return type per function is generally a good idea. I like the idea with the extra function, thank you!

mala/network/trainer.py
RandomDefaultUser: Alright, got it. I think I didn't look into the actual functionality deep enough, just noticed that code that had been authored by Josh had been deleted and just wanted to know if it was by accident or intentional. Your explanation makes sense, I am OK with deleting this part.

RandomDefaultUser (approved):
Thanks for incorporating all the changes and feedback, this looks great to me now and can be merged from my side!

PR #561, opened 2024-07-26
MALA for 2D materials

RandomDefaultUser:
This PR allows to perform TEM calculations (and, eventually, LAMMPS runs) for systems calculated using the 2D option of Quantum ESPRESSO ("assume_isolated = '2D'"). Basic calculations are already working, but there are currently still a few limitations:

- [ ] There is the issue of the general one-electron contribution bug, which also affects this PR (and generally non-metallic systems?)
- [ ] I haven't tested 2D systems with Gaussian descriptors or bispectrum calculations, there may be small adjustments necessary there

### DIFF
--- a/external_modules/total_energy_module/total_energy.f90
+++ b/external_modules/total_energy_module/total_energy.f90
@@ -250,11 +250,14 @@ SUBROUTINE init_run_setup(calculate_eigts)
      CALL ggen( dfftp, gamma_only, at, bg, gcutm, ngm_g, ngm, &
        g, gg, mill, ig_l2g, gstart )
   END IF
+
+
+  IF (do_cutoff_2D) CALL cutoff_fact()
+
   !
   !  This seems to be needed by set_rhoc()
   !
   CALL gshells ( lmovecell )
-
   !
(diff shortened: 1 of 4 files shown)


RandomDefaultUser:
I fixed the one-electron contribution bug. 

There is no bug in MALA or the TEM. The problem lies with the DFT-SCF convergence. Apparently, for systems with regions of small density/LDOS, standard DFT SCF accuracy does not suffice. E.g., with a certain SCF accuracy your DFT calculation may be accurate down to essentially mikro-eV/atom, but still the LDOS/density/DOS computed from it will give errors of tens to hundreds of eV/atom when compared with KS values. An example for this is attached for bilayer graphene with 76 atoms and a 7x7x1 k-grid.

I had to lower the convergence threshold by three orders of magnitude for my graphene data set to get the mean error on the LDOS vs. KS comparison from 74.0 meV/atom to 1.2 meV/atom.

convergence_issue

I have also tested 2D materials with bispectrum descriptors, for now it seems to work. If there are fixes along the way, I will open a PR.

Issue #508, opened 2024-03-26
Code duplications in `predictor.py`


elcorto:
https://github.com/mala-project/mala/blob/9b0c4fc45dc93c9e7165302d9cb0d56eb1e5fb32/mala/network/predictor.py#L37-L40 in `__init__` is the same as https://github.com/mala-project/mala/blob/9b0c4fc45dc93c9e7165302d9cb0d56eb1e5fb32/mala/network/predictor.py#L65-L68 in `predict_from_qeout()`. Since the former is in the constructor, I think the latter can be skipped.

Also here https://github.com/mala-project/mala/blob/9b0c4fc45dc93c9e7165302d9cb0d56eb1e5fb32/mala/network/predictor.py#L105-L110 `self.data.grid_dimension = self.parameters.inference_data_grid` is already performed in `__init__` so the check could be changed to something like

```py
if np.prod(self.parameters.inference_data_grid) == 0:
    ... code from else clause here ...
```

Note: `<= 0` would be correct but I gess `np.prod(self.parameters.inference_data_grid)` is never < 0.

RandomDefaultUser:
Hi @elcorto thanks for bringing this to my attention, I will have a look! :)

RandomDefaultUser:
I looked into the second part of this issue, and do no longer think this part of the code is redundant. The additional check makes sure that if the user specified an inference grid AFTER loading of the predictor class, this change is taken into account. Since the predictor object is loaded and the parameters object is not created beforehand, if you want to use a different inference grid this is the only way (I think?).

The first part is fixed by https://github.com/mala-project/mala/pull/562, so I am closing this issue.

PR #562, opened 2024-07-29
Improve predictor module

elcorto:
Some cleanup / update in `network/predictor.py` that I'd like to move out of feature branches into develop, in preparation of a larger PR later which includes one feature only.

### DIFF
--- a/mala/network/predictor.py
+++ b/mala/network/predictor.py
@@ -191,7 +191,6 @@ class Predictor(Runner):
         self, snap_descriptors, local_data_size=None
     ):
         """Forward a scaled tensor of descriptors through the NN."""
-
         # Ensure the Network is on the correct device.
         # This line is necessary because GPU acceleration may have been
         # activated AFTER loading a model.


mala/network/predictor.py
```diff
@@ -59,13 +59,6 @@ def predict_from_qeout(self, path_to_file, gather_ldos=False):
         predicted_ldos : numpy.array
             Precicted LDOS for these atomic positions.
         """
-        self.data.grid_dimension = self.parameters.inference_data_grid
-        self.data.grid_size = (
-            self.data.grid_dimension[0]
-            * self.data.grid_dimension[1]
-            * self.data.grid_dimension[2]
-        )
-
```
RandomDefaultUser, line 68: Was this part of the code redundant? If I am not mistaken the grid dimension is relatively crucial for the calculation, but if deleting this does not break the class itself there must be some redundancy then?

mala/network/predictor.py
```diff
@@ -198,6 +191,10 @@ def _forward_snap_descriptors(
         self, snap_descriptors, local_data_size=None
     ):
         """Forward a scaled tensor of descriptors through the NN."""
+        assert isinstance(
+            snap_descriptors, torch.Tensor
+        ), "snap_descriptors is not a Tensor"
+
```
RandomDefaultUser, line 197: Checking this via an assertion seems like a good idea generally, but just to clarify from my side - is it to be expected that a mismatch here happens often?

RandomDefaultUser (commented):
Hi @elcorto thanks for the improvements, I left a two comments for clarification, apart from that this looks good to me!

mala/network/predictor.py
elcorto: This is a code duplication, it fixes part of #508 .

mala/network/predictor.py
elcorto: I don't quite recall why I added that. I guess at some point I provided a numpy array which gave an error. I probably was confused about the fact that some methods expect a Tensor but return a numpy array, which leads users to the assumption that they can provide either.

Anyway, I'm fine with skipping that, also since it is a bit random (there are not too many similar checks in the rest of the code).

RandomDefaultUser: Got it, yeah I would suggest skipping it for now, maybe at some later point we should introduce such checks in general.

RandomDefaultUser:
I was wondering, should we address the second part of https://github.com/mala-project/mala/issues/508 in this PR as well? As you pointed out there, modifying the check should not be difficult.

mala/network/predictor.py
elcorto: OK agreed, I will push a change to this branch then.

elcorto:
> I was wondering, should we address the second part of #508 in this PR as well? As you pointed out there, modifying the check should not be difficult.

Ideally yes, but I left it for discussion and did not change anything since I was unsure how to change the check.

mala/network/predictor.py
elcorto: Changed in 8c78a62a

RandomDefaultUser:
> > I was wondering, should we address the second part of #508 in this PR as well? As you pointed out there, modifying the check should not be difficult.
> 
> Ideally yes, but I left it for discussion and did not change anything since I was unsure how to change the check.

Got it, I actually looked into it and do no longer think this part of the code is redundant. The additional check makes sure that if the user specified an inference grid AFTER loading of the predictor class, this change is taken into account. Since the predictor object is loaded and the parameters object is not created beforehand, if you want to use a different inference grid this is the only way (I think?). The part in `predict_from_qeout` is still redundant, since the check should (and does) happen in `predict_from_atoms`, so this PR is still good to go from my side, if the CI runs for the docs. 
I would close #508 though afterwards, I can add this exact comment to the issue to document it.

elcorto:
> > > I was wondering, should we address the second part of #508 in this PR as well? As you pointed out there, modifying the check should not be difficult.
> > 
> > 
> > Ideally yes, but I left it for discussion and did not change anything since I was unsure how to change the check.
> 
> Got it, I actually looked into it and do no longer think this part of the code is redundant. The additional check makes sure that if the user specified an inference grid AFTER loading of the predictor class, this change is taken into account. Since the predictor object is loaded and the parameters object is not created beforehand, if you want to use a different inference grid this is the only way (I think?). The part in `predict_from_qeout` is still redundant, since the check should (and does) happen in `predict_from_atoms`, so this PR is still good to go from my side, if the CI runs for the docs. I would close #508 though afterwards, I can add this exact comment to the issue to document it.

OK thanks for the double check. Let's go ahead and close #508 then with this PR.

RandomDefaultUser:
Sounds good, will merge here.

Issue #525, opened 2024-05-31
Use tempfile to handle LAMMPS and QE temporary files


RandomDefaultUser:
As pointed out by @elcorto we should use python's `tempfile` function (https://docs.python.org/3/library/tempfile.html to handle the temporary files created by the LAMMPS and QE interface, rather than coding everything up ourselves.

RandomDefaultUser:
PR #563 implements this for LAMMPS, only QE left to go.

RandomDefaultUser:
Closed by #595

PR #563, opened 2024-08-01
Use tempfile module for lammps file generation

timcallow:
This PR uses Python's `tempfile` module to generate the names for the temporary lammps files (#525). It should avoid the issues we had using the timestamp to generate unique names.

### DIFF
--- a/mala/descriptors/atomic_density.py
+++ b/mala/descriptors/atomic_density.py
@@ -137,9 +137,8 @@ class AtomicDensity(Descriptor):
         keep_logs = kwargs.get("keep_logs", False)
 
         lammps_format = "lammps-data"
-        self.lammps_temporary_input = os.path.join(
-            outdir, "lammps_input_" + self.calculation_timestamp + ".tmp"
-        )
+        self.setup_lammps_tmp_files("ggrid", outdir)
+
         ase.io.write(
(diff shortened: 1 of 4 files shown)


timcallow:
I think my `black` slightly reformatted a small section of existing code, maybe I'm using a different version and/or character limit?

I can revert that part to the original format if desired

elcorto:
> I think my `black` slightly reformatted a small section of existing code, maybe I'm using a different version and/or character limit?
> 
> I can revert that part to the original format if desired

We have the "recommended" version in [`.pre-commit-config.yaml`](https://github.com/mala-project/mala/blob/0fed9cc707979d2cdc0227c7a3eed4fd1fbbd291/.pre-commit-config.yaml) but we don't enforce that, e.g. in the CI I think.

If you format with `pre-commit` and run into the issue described in #568 then remove the line `language_version: python3.12`. I have this removed in a dev branch but haven't gotten around to submit a PR.

RandomDefaultUser (approved):
Thanks for implementing this @timcallow, looks great! Don't worry about the formatting, since we don't really enforce a format in the CI (yet, as @e
 pointed out) slight inconsistencies may happen, but are not a problem. The PR is ready to merge from my side.

Issue #564, opened 2024-08-02
Shuffling can cause (tiny) information loss


timcallow:
I was looking into issue #509 and noticed a slightly weird feature of the data shuffling function. 

Essentially, snapshots are created such that the **total** number of grid points (summed across snapshots) divided by the number of shuffled snapshots is an integer (lines 524-546 of `data_shuffler.py`). The dimensions of the shuffled snapshots is based on this.

However, when shuffling is done, the snapshots are populated by taking the number of grid-points **per snapshot** divided by the number of shuffled snapshots (lines 157-179). This is not necessarily an integer: for example, consider taking 3 snapshots from a grid of 200x200x200. 

The result of this is that, if the number of mixed snapshots is not a divisor for **each** of the original snapshot gridsizes, then some of the vectors in the final shuffled data (a very small number) will be completely zero. Here is one such example:

![Screenshot from 2024-08-02 16-11-38](https://github.com/user-attachments/assets/bf6828cf-845d-453f-b0c2-4d645e6b0782)

I very much doubt that a handful of zero vectors, given the typical grid sizes we work with, will affect the neural network training. But I have a couple of questions:

- Is this a known feature?
- Is it necessary to take exactly 1/nth from each original snapshot into the final mixed snapshots? Given the grid sizes we work with, there would be statistically no difference to first concatenate the input vectors into a single big one, and then simply shuffle that.

In my opinion, it would be better to do as described above. It simplifies the code and ensures there are no arbitrary zero vectors in the final training set. It would make the solution to #509 easier. 

What do you think @RandomDefaultUser? Am I missing something here?

Issue #509, opened 2024-04-04
Issue creating an arbitrary number of snapshots


timcallow:
When one tries to create an arbitrary number of snapshots using the `number_of_shuffled__snapshots` parameter in `data_shuffler.shuffle_snapshots`, an exception is raised if the number of snapshots requested is not an exact divisor of the number of data points. 

This means the burden is on the user to find an exact divisor, which is particularly inconvenient when one has a weird number of total grid points, which is quite common when mixing different mass densities. Since it isn't necessary for the snapshots to have exactly the same number of grid points, I would suggest to go with one of the following solutions:

1. The "last" snapshot ends up with extra or fewer grid points
2. All snapshots have the same number of grid points, any excess data is ignored (ok if lots of snapshots are requested, probably not so good if only a small number requested)
3. The closest number to the one requested that is also an exact divisor is sought

Imo, solution 1 is easiest and also best. The difference in the snapshot size will be very minor, considering the number of grid points we typically deal with. But @RandomDefaultUser let me know your thoughts. I can open a PR to implement a fix if you agree this is something to be fixed

RandomDefaultUser:
Thanks for bringing this to my attention! I think this is just an artifact of back when all snapshots in a training set had to have the same dimensions. I think at some point I also had something like 3 cooked up, but that was probably before you made the adjustments for multiple grid sizes during the mass density project.

I agree that option 1 sounds like the best one. If you want to take care of this and open a PR, I would highly appreciate that! :)

RandomDefaultUser:
Closed by #570.

PR #570, opened 2024-08-14
Flexible snapshot number for data shuffling

timcallow:
This PR allows the user to specify an arbitrary number of snapshots during the data shuffling procedure, even if the snapshot number isn't an exact divisor of the number of grid points in each snapshot.

In the non-exact-divisor case, a very small number of data points is "left out". To give an idea what very small means, consider a typical grid of 200x200x200 points, and a user requesting up to 10,000 snapshots. From the graph below, we can see that no more than ~0.1% of the initial data is left out in any of these choices.

Screenshot from 2024-08-14 13-10-54

This PR is a fix for issue #509.

Right now, I've implemented this only for the numpy case, so it's not possible with openPMD. Frankly I've no idea how to do it for the latter. I don't know how many users are currently using openPMD AND want to use an arbitrary snapshot number, if the answer is ~0 then we can just keep this as a todo for the future.

As mentioned in issue #564, I think the whole shuffling procedure could be simplified. But this PR ignores that possibility and integrates the new functionality into the code without making other changes.

### DIFF
--- a/mala/datahandling/data_shuffler.py
+++ b/mala/datahandling/data_shuffler.py
@@ -130,11 +130,33 @@ class DataShuffler(DataHandlerBase):
                 )
             )
 
+            # if the number of new snapshots is not a divisor of the grid size
+            # then we have to trim the original snapshots to size
+            # the indicies to be removed are selected at random
+            if self.data_points_to_remove is not None:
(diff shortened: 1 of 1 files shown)


RandomDefaultUser (approved):
Looks good to me, thanks @timcallow !

PR #574, opened 2024-08-27
Align LDOS vectors to a reference

timcallow:
Aligns a set of LDOS vectors to a reference. Optionally, the LDOS vectors can also be truncated from the left and right. This makes learning easier when the LDOS vectors become non-zero at different points, and/or are truncated (due to finite number of bands) at different points, such as when different mass densities are considered.

Remaining tasks need to be done or at least discussed before merging:

- [x] Parallelize over snapshots
- [x] Add alignment test
- [ ] ~~Extend to snapshots stored as openPMD~~
- [ ] ~~Include additional information files~~

### DIFF
--- a/examples/advanced/ex09_align_ldos.py
+++ b/examples/advanced/ex09_align_ldos.py
@@ -1,0 +1,32 @@
+import os
+
+import mala
+
+from mala.datahandling.data_repo import data_path
+
+"""
+Shows how to align the energy spaces of different LDOS vectors to a reference.
+This is useful when the band energy spectrum starts at different values, e.g.
+when MALA is trained for snapshots of different mass densities.
+
+Note that this example is only a proof-of-principle, because the alignment
+algorithm has no effect on the Be test snapshots (apart from truncation).
+"""
+
+
+parameters = mala.Parameters()
+parameters.targets.ldos_gridoffset_ev = -5
+parameters.targets.ldos_gridsize = 11
+parameters.targets.ldos_gridspacing_ev = 2.5
+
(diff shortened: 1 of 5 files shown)


timcallow:
Marking this as ready for review now. Some comments:

- The example is not that informative since the LDOS vectors in the test data repo are aligned (by our definition) anyway. If something better is required, I could either (i) Create additional npy files in the test data (seems overkill for one, likely rarely-used, example) or (ii) create new data OTF in the script, which might make the example a bit convoluted and detract from what's actually going on. Perhaps an example is not even required and it should be converted to a test, in which case (ii) would be fine. 
- I haven't extended to openPMD because I'm not familiar with it. This problem is likely to be a recurrent one (also happened in PR #570). Is the idea that all MALA developers understand how to manipulate openPMD data? Or one person (@RandomDefaultUser or @franzpoeschel) is responsible for implementing this functionality when required?
- There is an optional `number_of_electrons` parameter which could be removed if the calculation output file is included. This tells the user how much the energy is shifted from the QE reference energy. I decided it was easier just to have the `number_of_electrons` parameter than have to set up an LDOS object using the calculation outputs, but can change if desired.

franzpoeschel:
> * Or one person (@RandomDefaultUser or @franzpoeschel) is responsible for implementing this functionality when required?

I'll take a stab at it once I find the time. Feel free to merge before that, this can also go in separately.

RandomDefaultUser:
Thanks for the work @timcallow , looks good to me. I will merge this for now and open an issue to track the OpenPMD progress.

PR #582, opened 2024-10-08
Update GPU/production run docs

### LINKED ISSUE #553: Document `parameters.descriptor.use_atomic_density_energy_formula`
This setting can be used to speed up calculations with the TEM, but is not documented.

RandomDefaultUser:
Will close #553 .

### DIFF
--- a/docs/source/advanced_usage/predictions.rst
+++ b/docs/source/advanced_usage/predictions.rst
@@ -42,8 +42,8 @@ Likewise, you can adjust the inference temperature via
 
 .. _production_gpu:
 
-Predictions on GPU
-*******************
+Predictions on multiple GPUs
+****************************
 
 MALA predictions can be run entirely on a GPU. For the NN part of the workflow,
 this seems like a trivial statement, but the GPU acceleration extends to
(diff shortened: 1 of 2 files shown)


elcorto:
The text looks good, thanks. 

I have noticed that this (and probably other) settings in `Parameter` subclasses don't show up in the Sphinx API docs if they are not added as a property with a doc string (such as the new [`validation_metrics`](https://github.com/mala-project/mala/blob/develop/mala/common/parameters.py#L747). 

If this is correct and you think this is useful, then I can also add this -- the PR has "Maintainers are allowed to edit this pull request." set but I'm not sure whether I have the rights to push to your fork's branch. If not then you'd need to push this branch to mala proper first.

elcorto (commented):
The text looks good, thanks. 

I have noticed that this (and probably other) settings in `Parameter` subclasses don't show up in the Sphinx API docs if they are not added as a property with a doc string (such as the new [`validation_metrics`](https://github.com/mala-project/mala/blob/develop/mala/common/parameters.py#L747). 

If this is correct and you think this is useful, then I can also add this -- the PR has "Maintainers are allowed to edit this pull request." set but I'm not sure whether I have the rights to push to your fork's branch. If not then you'd need to push this branch to mala proper first.

RandomDefaultUser:
I added a docstring for this option, and opened this issue #587 because your comment made me realize there may be more than one docstring missing. 
I also opened #586 because I realized it should probably be moved to a different place altogether. This does not affect this PR specifically, but should be noted. 

Let me know if the PR is OK as is with these changes.

Issue #576, opened 2024-09-03
LDOS default units inconsitency


timcallow:
In mala's `DataConverter` class, in which the LDOS is processed from QE outputs, the default `target_units=None` in the `add_snapshot` function. This means that MALA performs **no** unit conversion by default on the QE outputs, which are in atomic units. 

On the other hand, the default units in MALA are eV and Angstrom. For example, in the `read_from_numpy_file` function of of the `LDOS` class, the default units `units="1/(eV*A^3)"`.

So if a user runs a QE calculation, processes the output from that using the `DataConverter` class, and then performs calculations on the LDOS using the `LDOS` class, there will be a units inconsistency and energies will be wrong. 

Although it's possible to override the default units in either the `DataConverter` or `LDOS` classes, I strongly think that, if a user 'does nothing', i.e. uses everywhere defaults, then these should align to produce a consistent result. I would propose to account for this in the `DataConverter` class by making the default `target_units="1/(Ry*Bohr^3)"`.

What do you think @RandomDefaultUser? If you agree I'll make a PR to address this issue.

RandomDefaultUser:
Thanks for pointing this out @timcallow, you are absolutely right, the default settings should agree. 
I am wondering though if instead of changing the default unit, we should implement an internal check mapping the unit to the specified target file. I.e., if it is from a cube file, we can (at least at the moment) safely assume it originated from QE. And therefore automatically assign the correct units. At any rate, feel free to create a PR to address the issue, that would be greatly appreciated!

PR #591, opened 2024-10-23
Change default units when reading from cube files

timcallow:
This PR addresses issue #576.

According to the Gaussian [documentation](https://gaussian.com/cubegen/), cube files are almost always in atomic units, and this is certainly the units used by QE, so it should be safe to make this assumption. 

I simply changed the default units for the `read_from_cube` functions. 

I also added a warning statement if users explicitly request the wrong units.

### DIFF
--- a/mala/targets/density.py
+++ b/mala/targets/density.py
@@ -102,7 +102,7 @@ class Density(Target):
         return return_dos
 
     @classmethod
-    def from_cube_file(cls, params, path, units="1/A^3"):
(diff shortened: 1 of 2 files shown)


timcallow:
Marked as draft because it's not behaving as expected

timcallow:
Ok this now works as expected @RandomDefaultUser

RandomDefaultUser (approved):
Looks great from my side, thank you! I will merge it.

Issue #592, opened 2024-10-23
Add coverage to tests


timcallow:
I suggest to add coverage to the tests. At the moment, if someone adds a new feature, the onus is on them to remember to add tests to check that it's working. Also it gives more confidence when making changes if you can see the test-suite is really covering the full range of functionality.

We could set it up so that if the coverage drops below a certain threshold, then the tests fail. This is how I set it up in atoMEC.

What do you think @RandomDefaultUser?

RandomDefaultUser:
I think that sounds like a great idea!

If you have the capacity, feel free to prepare a PR!

PR #594, opened 2024-10-25
Add coverage to tests

timcallow:
Addressing issue #592 

Since the coverage reports a number of 62%, I've set the `--cov-fail-under` parameter to 60%, i.e. the CPU tests will fail if the coverage dips below that.

We could use a third-party app like [codecov](https://about.codecov.io/) if we want more functionality - for example then we could have a separate test for coverage. That's what I did for atoMEC. This could also be introduced in a future PR if considered useful. I'll leave it up to you @RandomDefaultUser

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -223,7 +223,7 @@ jobs:
 
       - name: Test mala
         shell: 'bash -c "docker exec -i mala-cpu bash < {0}"'
-        run: MALA_DATA_REPO=$(pwd)/mala_data pytest --cov=mala -m "not examples" --disable-warnings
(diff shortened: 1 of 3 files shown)


RandomDefaultUser:
Hi @timcallow, thanks for the PR, looks great! 

I would merge it as is from the code side, but have a few clarifying questions about the process as a whole: 

1. Does a threshold of 60% mean we should add more tests? It probably does. 
2. What would be the advantages of a third party app? Are there disadvantages? At first glance it sounds like a good idea, and I would be very open to it! Feel free to answer these questions about the third party app in a new issue to document it.

RandomDefaultUser (approved):

PR #598, opened 2024-10-25
Address incosistencies in DataScaling infrastructure

### LINKED ISSUE #482: Data scaling: row vs. column and naming of methods
The `DataScaler` docs say (from `mala/datahandling/data_scaler.py`)

```

    Parameters
    ----------
    typestring :  string
        Specifies how scaling should be performed.
        Options:

        - "None": No normalization is applied.
        - "standard": Standardization (Scale to mean 0,
          standard deviation 1)
        - "normal": Min-Max scaling (Scale to be in range 0...1)
        - "feature-wise-standard": Row Standardization (Scale to mean 0,
          standard deviation 1)
        - "feature-wise-normal": Row Min-Max scaling (Scale to be in range
          0...1)
```

1. The "Row" parts of `feature-wise-*` suggest that inputs $X$ (e.g. bispectrum descriptors) and outputs $Y$ (LDOS) have shape `(n_features, n_samples)`, while the data in https://github.com/mala-project/test-data has `(18, 18, 27, 94)` ($X$) and `(18, 18, 27, 11)` ($Y$) and is, in `test/scaling_test.py`, reshaped to `(18*18*27, n_features)`. Further, the `DataScaler` code does `torch.mean(unscaled, 0, ...)`, so operates on columns (features, axis 0).

2. It should be clearly stated that the "normal" and "standard" modes operate on *the whole array*, rather than along the opposite axis of the one that the `feature-wise-` modes use, which users may assume.

3. Maybe "normal" should be renamed to something like "minmax", since ["Normalization" is usually something different](https://scikit-learn.org/stable/modules/preprocessing.html#normalization).

### LINKED ISSUE #483: Data scaling: API and documentation
From `mala/datahandling/data_scaler.py`:

`DataScaler.transform()`

```
    def transform(self, unscaled):
        """
        Transform data from unscaled to scaled.

        Unscaled means real world data, scaled means data as is used in
        the network. Data is transformed in-place.

        Parameters
        ----------
        unscaled : torch.Tensor
            Real world data.

        Returns
        -------
        scaled : torch.Tensor
            Scaled data.
        """
```

That method in-place modifies the array `unscaled` (that is documented), but it returns nothing (so `None`), so the doc string needs to be adapted (remove `Returns` section).

The method `DataScaler.inverse_transform()` returns a new array and doesn't modify the `scaled` input array.

The naming of the methods suggests that `DataScaler` operates like e.g., [`sklearn.preprocessing.StandardScaler`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html), but this API is different in the following ways:

* methods *always* return a transformed array
* methods have a `copy` arg that allows to select whether in-place mods are desired, see [`transform()`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler.transform) and [`inverse_transform()`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler.inverse_transform)
* the equivalent of `DataScaler`'s `start_incremental_fitting()` + `incremental_fit()` + `finish_incremental_fitting()` is probably targeting the functionality of [`StandardScaler.partial_fit()`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler.partial_fit) but seems more complex?
* Scaling always operates on feature columns and $X$ is assumed to have shape `(n_samples, n_features)` (see also #482)

RandomDefaultUser:
This PR will close #482 and should also close #483.

- [x] Address #482, i.e., fix docstrings and rename "normal" to "minmax"
- [x] Address #483, i.e., make API consistent with `sklearn` scalers


RandomDefaultUser:
@elcorto I hope I did not miss anything, but I think this should address the issues you have raised. I would appreciate your opinion/review.

elcorto:
It seems like you are still working on this PR? If so, please let me know when it is ready for a review.

elcorto:
> > It seems like you are still working on this PR? If so, please let me know when it is ready for a review.
> 
> As of today I was actually finished with the PR from my side, except of course if I missed something you wanted implemented?

I saw recent commits so I wanted to wait with the review, but OK then I'll look into it now.

docs/source/basic_usage/trainingmodel.rst
```diff
@@ -43,15 +43,22 @@ sub-objects dealing with the individual aspects of the workflow. In the first
...
+  applied to the entire array.
 
-* ``normal``: Min-Max scaling (Scale to be in range 0...1)
+* ``minmax``: Min-Max scaling (Scale to be in range 0...1) is applied to the entire array.
 
-* ``feature-wise-standard``: Row Standardization (Scale to mean 0, standard deviation 1)
+* ``feature-wise-standard``: Standardization (Scale to mean 0, standard
+  deviation 1) is applied to each feature dimension individually. I.e., if your
+  training data has dimensions (x,y,z,f), then each of the f rows with (x,y,z)
+  entries is scaled indiviually.
```
elcorto, line 56: As mentioned in #482, `DataScaler` silently assumes a 2D array of shape `(x * y * x, f)` since it does `torch.mean(unscaled, 0, ...)`, so unless I'm overlooking something obvious, I think this explanation is actually misleading.

Consider this:

```py
from sklearn.preprocessing import StandardScaler
from mala.datahandling.data_scaler import DataScaler
import torch as T
import einops
import numpy as np
from icecream import ic

reshape_to_2d = lambda x: einops.rearrange(x, "x y z f -> (x y z) f")
T.set_default_dtype(T.float64)


# sklearn's StandardScaler uses a biased estimator np.std(..., ddof=0), while
# torch uses T.std(..., correction=1) and there is no way to change this in
# StandardScaler, so we roll our own.
class MyStandardScaler:
    @staticmethod
    def fit_transform(X):
        assert X.ndim == 2
        x_mean = X.mean(0)[None, :]
        x_std = np.std(X.numpy(), axis=0, ddof=1)[None, :]
        return (X - x_mean) / x_std


with T.no_grad():
    arr_4d = T.rand(2, 3, 4, 5)
    arr_2d = reshape_to_2d(arr_4d)

    ##arr_2d_scaled_skl = StandardScaler().fit_transform(arr_2d)
    arr_2d_scaled_skl = MyStandardScaler().fit_transform(arr_2d)

    scaler_mala = DataScaler("feature-wise-standard")
    arr_2d_scaled_mala = arr_2d.clone()
    scaler_mala.fit(arr_2d_scaled_mala)
    # in-place mod!
    scaler_mala.transform(arr_2d_scaled_mala)

    ic(arr_2d_scaled_skl[:5, :])
    ic(arr_2d_scaled_mala[:5, :])
    assert np.allclose(arr_2d_scaled_mala, arr_2d_scaled_skl, atol=1e-15)

    # ValueError: Found array with dim 4. StandardScaler expected <= 2.
    ##arr_2d_scaled_skl = StandardScaler().fit_transform(arr_2d)

    # This works but shouldn't
    scaler_mala = DataScaler("feature-wise-standard")
    arr_4d_scaled_mala = arr_4d.clone()
    scaler_mala.fit(arr_4d_scaled_mala)
    # in-place mod!
    scaler_mala.transform(arr_4d_scaled_mala)

    arr_2d_from_4d_scaled_mala = reshape_to_2d(arr_4d_scaled_mala)
    assert not np.allclose(
        arr_2d_from_4d_scaled_mala, arr_2d_scaled_skl, atol=1e-1
    )
    ic(arr_2d_from_4d_scaled_mala[:5, :])
```

results in:

```
ic| arr_2d_scaled_skl[:5, :]: tensor([[-0.8424, -0.3500, -1.3310,  0.6888, -0.2759],
                                      [-0.1266, -1.1469, -0.2332, -0.3831, -0.7441],
                                      [-1.4691,  1.4810,  0.2890,  0.3123, -1.2043],
                                      [ 1.5169, -0.5278, -1.4609,  2.1362,  1.1646],
                                      [-0.6087,  0.2314,  0.9764, -1.0315,  0.6824]])
ic| arr_2d_scaled_mala[:5, :]: tensor([[-0.8424, -0.3500, -1.3310,  0.6888, -0.2759],
                                       [-0.1266, -1.1469, -0.2332, -0.3831, -0.7441],
                                       [-1.4691,  1.4810,  0.2890,  0.3123, -1.2043],
                                       [ 1.5169, -0.5278, -1.4609,  2.1362,  1.1646],
                                       [-0.6087,  0.2314,  0.9764, -1.0315,  0.6824]])
ic| arr_2d_from_4d_scaled_mala[:5, :]: tensor([[ 0.7071, -0.7071, -0.7071,  0.7071,  0.7071],
                                               [-0.7071, -0.7071, -0.7071,  0.7071, -0.7071],
                                               [-0.7071,  0.7071, -0.7071, -0.7071, -0.7071],
                                               [ 0.7071, -0.7071, -0.7071,  0.7071,  0.7071],
                                               [ 0.7071,  0.7071,  0.7071, -0.7071, -0.7071]])
```

So the code doesn't fail when given a 4D array but produces wrong results. Therefore, all docs mentioning `(x,y,z,f),` should be adapted. Also an array dimension check is probably a good idea.

mala/datahandling/data_scaler.py
```diff
@@ -295,6 +316,8 @@ def transform(self, unscaled):
 
         # Perform the actual scaling, but use no_grad to make sure
         # that the next couple of iterations stay untracked.
+        scaled = unscaled if copy is False else unscaled.clone()
```
elcorto, line 319:
```suggestion
        scaled = unscaled.clone() if copy else unscaled
```
is maybe a bit more pythonic

mala/datahandling/data_scaler.py
```diff
@@ -337,17 +362,29 @@ def inverse_transform(self, scaled, as_numpy=False):
...
             Real world data.
 
         """
+        # Backward compatability.
+        if not hasattr(self, "scale_minmax") and hasattr(self, "scale_normal"):
+            self.scale_minmax = self.scale_normal
+
+        # Perform the actual scaling, but use no_grad to make sure
+        # that the next couple of iterations stay untracked.
+        unscaled = scaled if copy is False else scaled.clone()
```
elcorto, line 383:
```suggestion
        unscaled = scaled.clone() if copy else scaled
```

docs/source/basic_usage/trainingmodel.rst
```diff
@@ -43,15 +43,22 @@ sub-objects dealing with the individual aspects of the workflow. In the first
...
 
-* ``feature-wise-normal``: Row Min-Max scaling (Scale to be in range 0...1)
+* ``feature-wise-minmax``: Min-Max scaling (Scale to be in range 0...1) is
+  applied to each feature dimension individually. I.e., if your training data
+  has dimensions (x,y,z,f), then each of the f rows with (x,y,z) entries is
+  scaled indiviually.
```
elcorto, line 61: See above.

mala/common/parameters.py
```diff
@@ -573,27 +573,45 @@ class ParametersData(ParametersBase):
...
+              to the entire array.
+            - "feature-wise-standard": Standardization (Scale to mean 0,
+              standard deviation 1) is applied to each feature dimension
+              individually. I.e., if your training data has dimensions
+              (x,y,z,f), then each of the f rows with (x,y,z) entries is scaled
+              indiviually.
+            - "feature-wise-minmax": Row Min-Max scaling (Scale to be in range
+              0...1) is applied to each feature dimension individually.
+              I.e., if your training data has dimensions (x,y,z,f), then each
+              of the f rows with (x,y,z) entries is scaled indiviually.
```
elcorto, line 589: See above.

mala/common/parameters.py
```diff
@@ -573,27 +573,45 @@ class ParametersData(ParametersBase):
...
```
elcorto, line 611: See above.

mala/datahandling/data_scaler.py
```diff
@@ -20,14 +21,23 @@ class DataScaler:
...
+          to the entire array.
+        - "feature-wise-standard": Standardization (Scale to mean 0,
+          standard deviation 1) is applied to each feature dimension
+          individually.
+          I.e., if your training data has dimensions (x,y,z,f), then each
+          of the f rows with (x,y,z) entries is scaled indiviually.
+        - "feature-wise-minmax": Min-Max scaling (Scale to be in range
+          0...1) is applied to each feature dimension individually.
+          I.e., if your training data has dimensions (x,y,z,f), then each
+          of the f rows with (x,y,z) entries is scaled indiviually.
```
elcorto, line 37: See above.

elcorto (commented):
I have one large comment that I think should be discussed and two small changes, the rest looks good, thanks a lot.

mala/datahandling/data_scaler.py
RandomDefaultUser: Sounds good, thanks for the suggestion!

mala/datahandling/data_scaler.py
RandomDefaultUser: Sounds good, thanks for the suggestion!

docs/source/basic_usage/trainingmodel.rst
RandomDefaultUser: Aaaaaah, I see what you mean now, sorry for misunderstanding. I have adapted the documentation and added an array check. Let me know if there is still something missing!

RandomDefaultUser:
I fixed the pipeline @elcorto after addressing your comments, could you let me know if this PR looks good from your side now?

elcorto (changes requested):
Thanks for the doc update in `DataScaler` and the 2d array check.

Two things that I stumbled upon:

* There are still places (see https://github.com/mala-project/mala/pull/598/files) that mention `(x,y,z,f)` that you probably missed to catch, namely in `docs/source/basic_usage/trainingmodel.rst` and `mala/common/parameters.py`.

* The docs in `DataScaler` and `mala/common/parameters.py` are identical. Maybe just keep them in one place instead, so for instance in `parameters.py` link to :class:`~mala/datahandling/data_handler.DataScaler` ?

RandomDefaultUser:
Thanks for the feedback and suggestions @elcorto , I implemented the changes and corrected (x,y,z) to (d) in the places you mentioned. I would keep the full doc strings in both the Parameters and DataScaler classes though. I think they are there for different audiences: a regular user who applies a model does not necessarily interact with the DataScaler class, so they need to be in the Parameters class. If a developer however wants to adapt the DataScaler code the proper doc string should be in the right place. I will add a note that changes in DataScaler should be propagated to the Parameters class though.

elcorto (approved):
OK then we're all set, I guess. Thanks a lot!

PR #606, opened 2024-11-15
MALA for multiple elements 

RandomDefaultUser:
While some work in that respect has already been performed at the Sandia National Labs, MALA does not have full multi-element capabilities yet. This PR aims to make multi-element systems fully available within MALA. 
This includes:

- [x] Two elements (example: Barium Oxide)
  - [x] Bispectrum descriptor calculation (serial)
  - [x] Bispectrum descriptor calculation (parallel/GPU)
  - [x] Total energy evaluation
  - [x] Total energy evaluation (using the atomic density formula)
  - [x] Testing the full workflow (only that it runs, not that the predictions are good)
  - [x] Implement a way to adjust the weights through the regular MALA interface
    - [x] Make a plan to enable descriptor calculation on the fly for (bispectrum) descriptor optimization, because that could become pretty relevant
  - [x] Confirming that the models are also accurate
- [x] More then two elements (example: Barium Nitrate)
  - [x] Bispectrum descriptor calculation (serial)
  - [x] Bispectrum descriptor calculation (parallel/GPU)
  - [x] Total energy evaluation
  - [x] Total energy evaluation (using the atomic density formula)
  - [x] Testing the full workflow (only that it runs, not that the predictions are good)
  - [x] Confirming that the models are also accurate
- [x] Add tests! 

While not the same issue in and of itself, it would be prudent to also tackle core electrons/"isolated" electronic levels in the same PR. I.e., if one has a system with, e.g., electronic levels located at -14 to -12 eV (core electrons) and then the next level is at, e.g., -1 to 0 eV, than by default the LDOS contains a lot of zeros. This can be alleviated by

- [x] Implement an LDOS cutting that cuts out zeros in the middle of the LDOS
- [x] Test that the entire worklfow runs with that


RandomDefaultUser:
I just realized that the TEM will work by default, since there is only one pseudopotential path anyway. The atomic density/Gaussian equation of course has to still be tackled.

RandomDefaultUser:
I confirmed basic functionality for the two element case. The code changes so far have been minimal, but the accuracy is not yet great. That could just be because I am using a minimal example (https://github.com/mala-project/test-data/tree/multielement, Barium Oxide unit cell), or because the bispectrum descriptors need tweaking. I will investigate this further and look into the three element case.

RandomDefaultUser:
The (L)DOS splitting has been tested and I have merged the necessary code.

RandomDefaultUser:
Also on-the-fly descriptor calculation has been implemented (already in the main branch of MALA) and merged here as well.

RandomDefaultUser:
I confirmed that models can in principle be computed on copper-zirconium and some chrome vanadium oxide calculations. I added enough tests to ensure that data can be pre- and postprocessed - the training routine is the same, so we don't have to explicitly test for it. 
The only caveat is that currently, the GPU calculation of the descriptors only works if all element weights are 1.0. We will fix this in the future, for now I added a note and make sure any calculation violating this constraint breaks.

test/workflow_test.py
```diff
@@ -53,50 +99,132 @@ def test_preprocessing(self):
...
+        }
+
+        # Multiple elements only work with LAMMPS.
+        systems = (
+            ["Be", "BaO"]
+            if importlib.util.find_spec("lammps") is not None
+            else ["Be"]
+        )
+
+        for system in ["Be", "BaO"]:
```
RandomDefaultUser, line 130:
```suggestion
        for system in systems:
```

PR #607, opened 2024-11-15
Fixing tiny information loss in shuffling

RandomDefaultUser:
As detailed by @timcallow in https://github.com/mala-project/mala/issues/564, our current shuffling can lead to information loss. This occurs when one requests a number of snapshots to be shuffled such that the total number of data points may be divisible by the new number of snapshots, while the individual number of grid points per snapshot are not. 

This PR fixes this behavior by making the behavior implemented in https://github.com/mala-project/mala/pull/570 the default. I.e., snapshot shuffling now means:

(1. If the user does not provide a number for the snapshot files to be _shuffled into_, it is implicitly assumed that `number_after_shuffling = number_before_shuffling`)
2. It is checked whether the individual snapshot grid can be divided by this new number of snapshots
3. If not, the overall data size is reduced, and the algorithm implemented in https://github.com/mala-project/mala/pull/570 kicks in, discarding a tiny amount of information 
4. Data is shuffled in snapshot-like files with dimensions (x,1,1,feature), i.e., effectively a 1D-array compatible with the MALA workflow. 

This PR also fixes an issue with non-cubic cells and adds a test.

### DIFF
--- a/mala/datahandling/data_shuffler.py
+++ b/mala/datahandling/data_shuffler.py
@@ -604,13 +604,14 @@ class DataShuffler(DataHandlerBase):
                     "will be left out of the shuffled snapshots.",
                 )
 
-            shuffle_dimensions = [
-                int(number_of_data_points / number_of_new_snapshots),
-                1,
-                1,
-            ]
         else:
             raise Exception("Invalid snapshot type.")
+
+        shuffle_dimensions = [
+            int(number_of_data_points / number_of_new_snapshots),
+            1,
+            1,
+        ]
 
         printout(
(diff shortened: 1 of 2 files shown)


RandomDefaultUser:
This PR only tackles the numpy side of things, for OpenPMD it omits some of the logic. I will address this separately and have disabled one of the tests for now,

timcallow:
LGTM. Checked it for shuffling input snapshots with different grid sizes and for a range of requested snapshot numbers

timcallow (approved):

Issue #573, opened 2024-08-22
Monitor observables every `N` epochs


elcorto:
When using `during_training_metric`, the respective quantity is calculated in every epoch, which may be costly if `during_training_metric="total_energy"`.

When using shuffled snapshots, adding the required `calculation_output_file` as in

```py
data_handler.add_snapshot(
    "Be_snapshot_shuffled1.in.npy",
    data_path,
    "Be_snapshot_shuffled1.out.npy",
    data_path,
    "va",
    calculation_output_file=os.path.join( data_path, "Be_snapshot1.out"),
)
```

may not be valid since the reference data in `Be_snapshot1.out` doesn't match the validation data. I'm not sure what data is read from this file, so this may or may not be a problem, but in any case one must provide some file here, else we see `Exception: Could not guess type of additional calculation data provided to MALA.`. 

In addition, #571 and #572 make it hard to use the feature in production at the moment.

So, is there a way to do something like `examples/basic/ex02_test_network.py` every `N` epochs only, where one defines non-shuffled test snapshots plus reference data `calculation_output_file="/path/to/qe.out"`. This would be independent of the validation data (one could call it second validation data set) and save compute as well.

nerkulec:
@RandomDefaultUser Sure, I can integrate this, I already have this on my branch

RandomDefaultUser:
Great, thank you!

nerkulec:
I implemented it in #584 in a way that LDOS error is evaluated every epoch, while other metrics only every `parameters.running.validate_every_n_epochs` epochs.
I usually train on shuffled data, and use unshuffled data for validation. That eliminates the need for two validation sets. @elcorto does that also work for you?

elcorto:
Thanks @nerkulec for this addition. I took the liberty to re-open this issue such that we can discuss this (which should be quickly resolved).

So to make sure I understand: There are two new parameters:

* `parameters.running.validation_metrics` (added in #560 I think)
* `parameters.running.validate_every_n_epochs`

I was under the impression that all of those need to be added, in the `running` case, as a property to `common.parameters.ParametersRunning`, with a doc string such that they show up in the sphinx docs?

So as I understand #584:

* `validate_every_n_epochs` will make the val loss be calculated every `validate_every_n_epochs` epochs instead of every (`validate_every_n_epochs=1`)
[...]

PR #609, opened 2024-11-18
Fix missing parameter docstrings

nerkulec:
Included a few of missing docstrings and removed a few parameters which are not used

### DIFF
--- a/mala/common/parameters.py
+++ b/mala/common/parameters.py
@@ -903,7 +903,7 @@ class ParametersHyperparameterOptimization(ParametersBase):
               that _xxx is only so that optuna will differentiate between
               variables. No reordering is performed by MALA; the order
               depends on the order in the list. _xxx can be essentially
-              anything.use_graphs
+              anything.
 
         Users normally don't have to fill this list by hand, the hyperparamer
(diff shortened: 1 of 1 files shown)


elcorto (approved):
Thanks this looks good. I still think an extended example as suggested in #573 would be a nice-to-have, but as far as doc strings go, all good. :+1:

RandomDefaultUser:
Thanks for this PR @nerkulec, I have one questin before merging. I just finished working on #613 , and was looking at this PR as well so as not to do the same work twice. I realized that the following docstrings in `ParametersRunning` are still missing: 

```python
l1_regularization
l2_regularization
learning_rate_embedding
input_noise
batch_norm
dropout
use_compression
checkpoint_best_so_far
snapshots_per_epoch
training_log_interval
```

If I am not mistaken, you added these parameters. Could you also add a docstring for these parameters? If I am mistaken and some of these are from older MALA changes, e.g., by Josh, let me know and I will take care of them.

RandomDefaultUser:
I will merge this PR, and then adapt https://github.com/mala-project/mala/pull/613 accordingly.

Issue #608, opened 2024-11-15
Implement all the shuffling interface changes for OpenPMD as well


RandomDefaultUser:
There have been quite a few changes to the shuffling interface #509 #564, #570, #607. Currently, these have ignored the OpenPMD interface and the divergence is slowly creeping up on us (see disabled test in #607 ). We should adapt the OpenPMD interface to implement these changes as well.

PR #610, opened 2024-11-21
Quickfixing OpenPMD Shuffling

RandomDefaultUser:
The current shuffling algorithm is not fully adapted to OpenPMD yet. This PR removes some inconsistencies and adds a ValueError to make clear to users what is actually happening. Arbitrary number of snapshot shuffling capabilities for OpenPMD soon.


RandomDefaultUser:
See also https://github.com/mala-project/mala/issues/608

Issue #586, opened 2024-10-18
Move "use_atomic_density_energy_formula" parameter


RandomDefaultUser:
As a legacy setting, this parameter, which controls whether Gaussian descriptors are calculated for the evaluation of the total energy, lives in the Descriptor subclass of the `Parameters` class. This should be moved. I am opening a new issue for this, since moving may break some code.

RandomDefaultUser:
Fixed by #611

PR #611, opened 2024-11-21
Move the flag `use_atomic_density_energy_formula`

RandomDefaultUser:
As detailed in #586, it makes sense to move the parameter  `use_atomic_density_energy_formula` from the descriptor subclass to the `Parameters` object itself. The idea here is that ideally, users can use models without having to worry about the subclasses at all; so the top-level parameters give access to all necessary acceleration options. 

The PR is backwards compatible, old models can be loaded without issue. Enabling GPU and LAMMPS at the same time will further automatically activate this function.


RandomDefaultUser:
Side note, to clarify commit message: GPU usage has now also been tested

PR #616, opened 2024-11-28
v1.3.0 - Into the multi-GPU-niverse

RandomDefaultUser:
# New features

- Multi-GPU inference: Models can now make predictions on an arbitrary number of GPUs
- Multi-GPU training: Models can now be trained on an arbitrary number of GPUs
- MALA now works with 2D materials, i.e., any system which is only periodic in two dimensions
- Bispectrum descriptor calculation now possible in python
	- This is route is significantly slower than LAMMPS, but can be helpful for developers who want to test the entire MALA modeling workflow without installing LAMMPS
- Logging for network training has been overhauled and now allows for the logging of multiple metrics 

- (EXPERIMENTAL) Implementation of a mutual information based metric to replace/complement the ACSD	
- (EXPERIMENTAL) Implementation of a class for LDOS alignment to a reference energy value; this can be useful for models across multiple mass densities	

# Changes to API/user experience

- New parallelization parameters available:
	- ``use_lammps`` - enable/disable LAMMPS (enabled by default, recommended for optimal performance, will automatically be disabled if no LAMMPS is found on the machine)
	- ``use_atomic_density_formula`` - enable the use of total energy evaluation based on a Gaussian representation (enabled if LAMMPS and GPU are enabled, recommended for optimal performance, details can be found in our [paper on size transfer](https://www.nature.com/articles/s41524-023-01070-z))
	- ``use_ddp`` - enable/disable DDP, i.e., Pytorch's distributed training scheme (disabled by default)
- Multiple LAMMPS/QE calculations can now be run in one directory
	- Prior to this version, doing so would lead to problems due to the file based nature of these interfaces
	- This allows for multiple simultaneous inferences in the same folder
- Class ``SNAP`` and all associated options are deprecated, use ``Bispectrum`` and associated options instead
- Default units for reading from .cube files are now set to units commonly used within Quantum ESPRESSO, this should make it easier to avoid inconsistencies in data sampling
- ASE calculator ``MALA`` now reads models with ``load_run()`` instead of ``load_model`` which is more consistent with the rest of MALA
- Error reporting with the ``Tester`` class has been improved, all errors and energy values reported there are now consistently given in meV/atom
- MALA calculators (LDOS, density, DOS) now also read energy contributions and forces from Quantum ESPRESSO output files, these can be accessed via properties

# Fixes

- Updated various performance/accessibility issues of CI/CD
- Fixed compatability with newer Optuna versions
- Added missing docstrings
- Fixed shuffling interface, arbitrary numbers of shuffled snapshots can now be created without loss of information
- Fixed inconsistency of density dimensions when using directly from cube file
- Fixed error when using GPU graphs with arbitrary batch sizes


acangi:
Looks good to me. Fantastic job @RandomDefaultUser!

RandomDefaultUser:
The CI is currently failing because there is an "internal server error" being sent back by RODARE... I don't know why that is, but it is most likely only a temporary problem of RODARE. I will resubmit the CI later today, and if the error persists, on Monday. It it persists thereafter, I will contact RODARE staff.

srajama1:
Looks good to me, thank you @RandomDefaultUser !

srajama1 (approved):

PR #622, opened 2024-12-05
Add arbitrary shuffling for openPMD input

franzpoeschel:
Shuffling openPMD data so far worked only if the number of output snapshots cleanly divides the first dimension of the input data. This PR overcomes that restriction.

TODO:

- [x] Document the code better, some of the indexing is non-trivial


RandomDefaultUser (approved):
Looks good to me! There is one small caveat, which I think is inconsequential: The shuffled snapshots in the arbitrary mode are not identical for numpy and openpmd. I think that does not matter much, since we only care about randomizing data, not having the exact same randomization. But I wanted to mention it to document it for future reference.

PR #630, opened 2025-01-07
Implement on-the-fly descriptor calculation

RandomDefaultUser:
This PR gives MALA the capability to compute descriptors on the fly. In doing so, it allows for more refined hyperparameter optimization of descriptor related hyperparameters, since the training-free methods we initially envisioned may not (yet) be enough in many cases. The starting point for this on-the-fly methods are the .json files the DataConverter class can generate. 

Since this is a pretty fundamental change in the MALA pipeline, it requires thorough testing. Namely, the following capabilities have to be implemented:

- [x] Single CPU/GPU training:
   - [x] RAM based training
   - [x] RAM based training w/ checkpointing
   - [x] Lazy loading training
   - [x] Lazy loading training w/ checkpointing
   - [x] Lazy loading prefetch training
   - [x] Lazy loading prefetch training w/ checkpointing
- [ ] ~~DDP training:~~
   - [ ] ~~RAM based training~~
   - [ ] ~~RAM based training w/ checkpointing~~
   - [ ] ~~Lazy loading training~~
   - [ ] ~~Lazy loading training w/ checkpointing~~
- [x] Adapt Tester class
- [x] Write Example
- [x] Write Documentation
- [x] Rename "additional data" in the DataConverter class
- [x] Add Shuffling to this on-the-fly calculation
- [x] Integrate new functionality in test suite
- [x] Add an automatic detection of provided data


RandomDefaultUser:
I just looked into the DDP side of things. The problem here seems to be that DDP allocates the GPU for its own usage, so when LAMMPS wants to use them, they are not free. This should still be solvable, but requires more modification in the code. I will first address the other open issues and address this issue later on, maybe in a separate PR.

RandomDefaultUser:
Skipping DDP implementation for now since it is not that crucial at the moment and can  be added later. See https://github.com/mala-project/mala/issues/631

PR #644, opened 2025-03-03
ACE documentation updates

jmgoff:
Adding to updates from @RandomDefaultUser and including more thorough ACE documentation. Beginning with ace descriptor class. This isn't finished yet, but I wanted to get a draft open.

### DIFF
--- a/mala/common/parameters.py
+++ b/mala/common/parameters.py
@@ -471,50 +471,47 @@ class ParametersDescriptors(ParametersBase):
         descriptor calculation.
 
     ace_ranks : list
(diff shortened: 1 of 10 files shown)


jmgoff:
Now compatible with `pydocstyle`. Was able to successfully build sphinx docs with the most recent commit. While some naming conventions could be updated still to something more descriptive (marked with TODOs), I would call this finished for now.

RandomDefaultUser:
Closed through https://github.com/mala-project/mala/pull/649
(I know it would technically be better to get push access on this branch and update this PR, but since we are working on a bit of a time constraint I opted for this route.)

Issue #559, opened 2024-07-12
Problem building QE (total energy module)


elect86:
So, I've been trying to tackle this for a while now and I got so close, but I was stopped by one damn last issue, nonetheless

I'm on Ubuntu

I've been trying to follow the docs [here](https://mala-project.github.io/mala/install/installing_qe.html):

- QE 7.2
- `apt install openmpi-bin`

I had [these issues](https://gist.github.com/elect86/b2ab1794da8673a95496013908e7091d#file-gistfile1-txt-L703-L714), which I managed to fix by getting OpenMPI using these packages [here](https://webpages.charlotte.edu/abw/coit-grid01.uncc.edu/ParallelProgSoftware/Software/OpenMPIInstall.pdf)

`sudo apt-get install openmpi-bin openmpi-doc libopenmpi-dev`

I need to install, in addition, `gfortran` otherwise `./configure` won't see any compiler

One difference is that `gfortran` will install the last one, that is `gfortran-13`, while @RandomDefaultUser has a working scenario with 11, I tried to install that, but `./configure` doesn't detect it as well

Also, `gfortran` will actually install `gfortran-13 gfortran-13-x86-64-linux-gnu libgfortran-13-dev`

So, in order to match Lenz specs, I tried the same for 11, but

>  127 elect@5800x:~$ sudo apt install gfortran-11 gfortran-11-x86-64-linux-gnu libgfortran-11-dev
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
>
> The following packages have unmet dependencies:
 gfortran-11-x86-64-linux-gnu:i386 : Depends: gcc-11-x86-64-linux-gnu:i386 (= 11.4.0-9ubuntu1cross1) but it is not going to be installed
                                     Depends: libgfortran-11-dev-amd64-cross:i386 (>= 11) but it is not installable
E: Unable to correct problems, you have held broken packages.

Interesting, the very same error with 12

So, I gave up on that front and went forward with 13, which is fine under this point of view

Then, trying to compile, [these problems](https://gist.github.com/elect86/b02cf520025131cfe37f14dfd38b7773)

I managed to fix them by adding [these flags](https://mattermodeling.stackexchange.com/questions/12825/wannier90-compilation-error-mpi) into make.inc
```
#  Generic flag name:
[...]

PR #645, opened 2025-03-06
Update requirements

### LINKED ISSUE #643: mpi4py missing in requirements.txt
Is this by design? If so then this makes sense because installing it requires people to have an MPI implementation installed as well. It could be added to a `requirements_extra.txt` though, or when we switch to a full `pyproject.toml` then 

```toml
 dependencies = [
    "torch",
    "..."
]


[project.optional-dependencies]
test = [
    "pytest",
    "..."
]

extra = [
    "mpi4py",
    "..."
]
```

elcorto:
Closes #643

### DIFF
--- a/docs/source/install/installing_mala.rst
+++ b/docs/source/install/installing_mala.rst
@@ -7,11 +7,16 @@ Prerequisites
 MALA supports any Python version starting from ``3.10.4``. No upper limit on
 Python versions are enforced. The most recent *tested* version is ``3.10.12``.
 
-MALA requires ``torch`` in order to function. As the installation of torch
-depends highly on the architecture you are using, ``torch`` will not
-automatically be installed alongside MALA. Please obtain a suitable version
-of ``torch`` from the `official website <https://pytorch.org/>`_ before
-continuing.
+MALA requires ``torch`` (https://pytorch.org). Most install sites will want to
+use an already installed ``torch`` or install a specific version before
+installing the MALA package itself, for instance
+
+.. code-block:: sh
+
+    pip install --index-url https://download.pytorch.org/whl/cu<cuda_version> torch==<torch_version>
+
+If no ``torch`` is found, we install the latest GPU-enabled version via ``pip
+install torch`` (see ``requirements.txt``).
 
 Installing the Python library
 *****************************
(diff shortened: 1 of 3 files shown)


RandomDefaultUser (changes requested):
Hi @elcorto thanks for the PR and sorry that I didn't have time to take a look earlier. Code wise this all looks great, but I think we should mention the changes in the documentation, namely here:
https://github.com/mala-project/mala/blob/develop/docs/source/install/installing_mala.rst

docs/source/install/installing_mala.rst
```diff
@@ -7,11 +7,16 @@ Prerequisites
...
+MALA requires ``torch`` (https://pytorch.org). Most install sites will want to
+use an already installed ``torch`` or install a specific version before
+installing the MALA package itself, for instance
+
+.. code-block:: sh
+
+    pip install --index-url https://download.pytorch.org/whl/cu<cuda_version> torch==<torch_version>
+
+If no ``torch`` is found, we install the latest GPU-enabled version via ``pip
+install torch`` (see ``requirements.txt``).
```
RandomDefaultUser, line 19: Doesn't that read like the user has to do that manually, whereas with this PR this will be done be done automatically?

RandomDefaultUser (changes requested):

RandomDefaultUser:
I left one small comment, apart from that this looks good to merge - sorry for the delay.

docs/source/install/installing_mala.rst
elcorto: The intent was encourage people to install a version of torch that matches their hardware, but well people who are aware of the torch install procedure do that already, and those who aren't are probably fine if we `pip install torch` for them. I'll rephrase the section.

docs/source/install/installing_mala.rst
```diff
@@ -7,11 +7,16 @@ Prerequisites
 MALA supports any Python version starting from ``3.10.4``. No upper limit on
 Python versions are enforced. The most recent *tested* version is ``3.10.12``.
```
elcorto, line 8: I just saw that this is also outdated. [Here](https://github.com/mala-project/mala/issues/559#issuecomment-2231786678) we say we support up to Python 3.11, is this still true?

docs/source/install/installing_mala.rst
RandomDefaultUser: Thank you!

docs/source/install/installing_mala.rst
RandomDefaultUser: I think python 3.11. might work, but officially, we only support up to 3.10.4 now. Making sure MALA runs with newer python versions is high on my list of priorities, but for now, it has not been really tested.

RandomDefaultUser (approved):
This looks good to me now, thanks @elcorto !

PR #648, opened 2025-03-14
Overhaul activation list

### LINKED ISSUE #567: Make activation applied to the output layer optional
I saw @nerkulec's [branch](https://github.com/mala-project/mala/tree/no_activation_after_last_layer) which made me realize that by default, `FeedForwardNet` applies an activation to the output layer. This is somewhat unusual for regression (see e.g. [here](https://d2l.ai/chapter_multilayer-perceptrons/mlp.html#from-linear-to-nonlinear) eq. 5.1.3).

If this is done to enforce positive LDOS values, then one probably needs to fix this to `ReLU` only.

However, having any activation here (also `ReLU`) prevents using a MALA model together with a fast last layer code path in https://github.com/aleximmer/Laplace. There is [an alternative way to do the same](https://github.com/aleximmer/Laplace/issues/217) but this has not been benchmarked extensively for the MALA case so far.

In any case, I think such an activation should be opt-in (or -out). It should however not removed, in order to have backward compatibility with results already produced with the current code version. One could for instance introduce a bool setting `parameters.network.use_output_activation`. Then, the logic here

https://github.com/mala-project/mala/blob/0fed9cc707979d2cdc0227c7a3eed4fd1fbbd291/mala/network/network.py#L214-L224

needs to be adapted as well.

RandomDefaultUser:
Changes the behavior of layer_activations which currently has to be a list of either a single activation function or exactly enough to cover all layers (including the output layer). The goal of this PR is to make layer_activations a list which is applied verbatim - i.e., allowing to skip layers, leave layers without activation function, etc. and leave the option of having layer_activations being a singluar string for setting all to the same value. Part of this has been implemented for the graphene project. Will close #567 

Open to-dos:
- [x] Set a default - should setting layer_activations to a singular value apply the activation to all layers or all layers up until the output layer?
- [x] Implement a functionality to skip layers
- [x] Test/document


RandomDefaultUser:
I am personally leaning towards keeping the old default for now, and I actually have a maybe semi-good reason for it. Since we are usually using LeakyReLU/ReLU as activation function, this leaves positive values in the output layer unchanged, BUT it (almost/fully) zeroes out negative values, and the LDOS should always be zero. No we can (and do) enforce this positivity explicitly as well (see parameter `restrict_targets`), but this way it is explicitly coded into the network. 
But I am open to change that default if you @nerkulec @elcorto think. 

With this PR it is now definitely possible to build networks without an activation function in the last layer or even skip layers.

elcorto:
Oh, I didn't know about `restrict_targets` :)

I have no strong opinions re. the default. But I'd for sure appreciate to have the last layer activation be opt-out (if the old default is kept, IIRC @nerkulec as well). For the single activation case, I fixed this with a new parameter `params.ff_use_output_activation` like so:

```diff
commit 2be929aecb423ba450119ba142550345d38d36e8
Author: Steve Schmerler <git@elcorto.com>
Date:   Tue Aug 20 11:53:06 2024 +0200

    HOTFIX: opt-in ff_use_output_activation (#567)

    Don't fix multi-activation case

diff --git a/mala/network/network.py b/mala/network/network.py
--- a/mala/network/network.py
+++ b/mala/network/network.py
@@ -255,7 +255,9 @@ class FeedForwardNet(Network):
                     )
                 )
             )
-            if i < self.number_of_layers - 1:
+            if (
+                i < self.number_of_layers - 1
+            ) or self.params.ff_use_output_activation:
                 try:
                     if use_only_one_activation_type:
                         self.layers.append(
```

The patch probably doesn't apply to this PRs changes, but you get the idea.

For the multi activation case (list), using a list of N-1 should do the trick, right?

mala/common/parameters.py
```diff
@@ -344,16 +344,25 @@ class ParametersNetwork(ParametersBase):
...
+        will map them directly to the correct activation functions or
+        torch.nn.Module objects containing the activation functions directly
+        OR None, in which case no activation function is used. The None
+        can be ommitted at the end, but is useful when layers without
+        activation functions are to be added in the middle
+        Currently supported activation function strings are:
+
+            - "Sigmoid"
+            - "ReLU"
+            - "LeakyReLU" (default)
```
elcorto, line 365: `mala.network.network.Network._activation_mappings` also allows "Tanh", so this list needs to be updated. But what is the reasoning behind the restriction to a few functions? Why not

```py
>>> getattr(torch.nn, "ReLU")
torch.nn.modules.activation.ReLU

>>> getattr(torch.nn, "ReLU")()
ReLU()

>>> isinstance(getattr(torch.nn, "ReLU"), torch.nn.Module)
False

>>> isinstance(getattr(torch.nn, "ReLU")(), torch.nn.Module)
True
```

Then `_activation_mappings` could be dropped.

One could also allow people to pass any (list of) `nn.Module` instance(s), like `torch.nn.LeakyReLU(negative_slope=0.02)` instead of strings, which would allow passing arguments and even custom activations, but I'd probably not do this as part of this PR but when the need arises.

mala/network/network.py
```diff
@@ -296,6 +294,24 @@ def forward(self, inputs):
...
+            Activation function to be appended.
+        """
+        if activation_function is None:
+            pass
+        elif isinstance(activation_function, str):
+            self.layers.append(
+                self._activation_mappings[activation_function]()
+            )
+        elif isinstance(activation_function, nn.Module):
+            self.layers.append(activation_function)
```
elcorto, line 313: This branch seems to allow `activation_function` to be something like `torch.nn.modules.activation.ReLU()` in addition to a string? If so, then one should maybe catch the case that `activation_function = torch.nn.modules.activation.ReLU` (class instead of instance), which for which the `isinstance()` test is False. I'd probably throw an error in that case.

elcorto:
As for zeroing negative LDOS outputs via a last layer activation: isn't a global `LeakyReLU` default problematic? But having a `ReLU` on the last layer only and a `LeakyReLU` on all others doesn't work as a default, since this needs to be specified as a list `["LeakyReLU", ..., "LeakyReLU", "ReLU"]`, which depends on the architecture.

RandomDefaultUser:
> I have no strong opinions re. the default. But I'd for sure appreciate to have the last layer activation be opt-out (if the old default is kept, IIRC @nerkulec as well). For the single activation case, I fixed this with a new parameter `params.ff_use_output_activation` like so:

I see your point, but my intention with this PR was to limit the number of parameter settings so that the usage does not become too confusing. So instead of having this opt-out, I thought if one can simply build the list from scratch, all cases can be covered.  So

> For the multi activation case (list), using a list of N-1 should do the trick, right?

would then be exactly right. 

The default should then be the most common/most useful use case.

But I think you may be right, it could be useful to set only a single parameter and have that fill the entire list and have another parameter decide what happens to the last layer. I will look into this right away and update the PR.

mala/common/parameters.py
RandomDefaultUser: I modified the internal behavior - now any combination of strings, classes and objects (of course all have to be a subclass of torch.nn.Module) can now be supplied. I also added Tanh to the list.

mala/network/network.py
RandomDefaultUser: See above, both instances and classes are now allowed.

mala/common/parameters.py
elcorto: Ah great, thanks re. the option to pass instances (and even classes!).

The other observation I had was that if `activation_function` is a string, then one can always do `getattr(torch.nn, activation_function)()` in `_append_activation_function()`. Then the dict `self._activation_mappings` is not needed and the user is not restricted to the activations in there.

elcorto:
> > For the multi activation case (list), using a list of N-1 should do the trick, right?
> 
> would then be exactly right.
> 
> The default should then be the most common/most useful use case.
> 
> But I think you may be right, it could be useful to set only a single parameter and have that fill the entire list and have another parameter decide what happens to the last layer. I will look into this right away and update the PR.

Thanks for adding `layer_activations_include_output_layer`. The equivalent parameter I added above was merely a hot-fix. 
In the spirit of minimizing the number of settings, what could also work is the default being a last-layer activation if `parameters.network.layer_activations` is a single item that gets added to each layer. Then, if needed, one can switch the last layer activation off by specifying a N-1 length list of activations as you already stated. If I understand correctly, with `layer_activations_include_output_layer`, there are now two ways to do that? If so, and if it makes the code simpler, then we may consider having only one option.

elcorto:
> > As for zeroing negative LDOS outputs via a last layer activation: isn't a global `LeakyReLU` default problematic? But having a `ReLU` on the last layer only and a `LeakyReLU` on all others doesn't work as a default, since this needs to be specified as a list `["LeakyReLU", ..., "LeakyReLU", "ReLU"]`, which depends on the architecture.
> 
> I think you are in principle correct, it makes more sense to use ReLU rather than LeakyReLU for the last layer. With this PR I would first like to consolidate the layer activation interface while maintaining compatibility to older workflows. We could definitely implement a default functionality like this in the future though. Effectively, `restrict_targets` currently gives similar results.

Agreed, I didn't intend to suggest a new feature or default. Maybe it would be beneficial to mention `restrict_targets` briefly in the activations-related docs, so that users are aware of the options - either use `["LeakyReLU", ..., "LeakyReLU", "ReLU"]` or `restrict_targets`.

mala/common/parameters.py
RandomDefaultUser: Ups, sorry, didn't read your comment carefully enough. I added the `getattr` mechanism. I am using similar mechanisms in other places, so it makes sense to do that here as well. Thanks for the hint!

RandomDefaultUser:
> > > As for zeroing negative LDOS outputs via a last layer activation: isn't a global `LeakyReLU` default problematic? But having a `ReLU` on the last layer only and a `LeakyReLU` on all others doesn't work as a default, since this needs to be specified as a list `["LeakyReLU", ..., "LeakyReLU", "ReLU"]`, which depends on the architecture.
> > 
> > 
> > I think you are in principle correct, it makes more sense to use ReLU rather than LeakyReLU for the last layer. With this PR I would first like to consolidate the layer activation interface while maintaining compatibility to older workflows. We could definitely implement a default functionality like this in the future though. Effectively, `restrict_targets` currently gives similar results.
> 
> Agreed, I didn't intend to suggest a new feature or default. Maybe it would be beneficial to mention `restrict_targets` briefly in the activations-related docs, so that users are aware of the options - either use `["LeakyReLU", ..., "LeakyReLU", "ReLU"]` or `restrict_targets`.

Good point - I updated the docs accordingly. Does this suffice in your opinion?

elcorto (approved):
This looks very good now, thanks.

PR #651, opened 2025-03-27
Install readme: fix test-data tag name

brian-kelley:
It's "1.8.1", not "v1.8.1"


RandomDefaultUser:
Thanks for catching that - actually, at this point, it is already 1.9.0. Let me see if I can find a solution that automatically updates the docs, else wise these inconsistencies will always happen.

pcagas:
The version (without the `v`) has been updated to 1.9.0 in #663. From now on, this should be updated automatically with every `test-data` release; see https://github.com/mala-project/test-data/pull/15 for detail

PR #653, opened 2025-04-14
Align LDOS: Read from and write to openPMD

franzpoeschel:
Uses `read_from_openpmd_file` and `write_to_openpmd_file` on the target calculator to connect the LDOS Aligner to openPMD.


examples/advanced/ex09_align_ldos.py
```diff
@@ -30,3 +32,37 @@
...
+
+    parameters = mala.Parameters()
+    data_handler = mala.DataHandler(parameters)
+    for i in range(1, 4):
+        data_openpmd = data_handler.target_calculator.read_from_openpmd_file(
+            f"{data_path}/aligned/Be_snapshot0.out.h5")
+        data_numpy = data_handler.target_calculator.read_from_numpy_file(
+            f"{data_path}/aligned/Be_snapshot0.out.npy")
+        if not numpy.allclose(data_numpy, data_openpmd):
+            raise Exception("Inconsistency in snapshot", i)
```
franzpoeschel, line 66: Should this be moved to a test?

mala/datahandling/ldos_aligner.py
```diff
@@ -279,30 +299,34 @@ def align_ldos_to_ref(
...
-                ".npy", ".ldos_shift.info.json"
-            )
-
+            target_name = os.path.join(save_path, save_name)
             os.makedirs(save_path, exist_ok=True)
 
-            if "*" in save_name:
-                save_name = save_name.replace("*", str(idx))
-                ldos_shift_info_save_name.replace("*", str(idx))
+            self.ldos_parameters.ldos_gridsize = ldos_shifted.shape[-1]
```
franzpoeschel, line 305: This is needed for creation of correct dimensions in openPMD writing. The LDOS aligner apparently has some awareness of MPI. 
`left_index_ref` is computed by rank 0 and then distributed to others, `e_index_cut` however seems to be computed by everyone independently based on local data. I'm not sure if this can possibly yield consistent results, even with pure Numpy.
But in case this module actually supports MPI, then this is entirely untested for openPMD atm.

mala/datahandling/ldos_aligner.py
```diff
@@ -146,13 +144,24 @@ def align_ldos_to_ref(
...
-                mmap_mode="r",
-            )
+            if snapshot_ref.snapshot_type == 'numpy':
+                ldos_ref = np.load(
+                    os.path.join(
+                        snapshot_ref.output_npy_directory,
+                        snapshot_ref.output_npy_file,
+                    ),
+                    mmap_mode="r",
+                )
```
franzpoeschel, line 154: Might want to use `read_from_numpy_file` here for consistency instead of raw numpy calls.

mala/datahandling/ldos_aligner.py
```diff
@@ -205,13 +214,24 @@ def align_ldos_to_ref(
...
-                mmap_mode="r",
-            )
+            if snapshot_ref.snapshot_type == 'numpy':
+                ldos = np.load(
+                    os.path.join(
+                        snapshot.output_npy_directory,
+                        snapshot.output_npy_file,
+                    ),
+                    mmap_mode="r",
+                )
```
franzpoeschel, line 224: dito

examples/advanced/ex09_align_ldos.py
RandomDefaultUser: I think that would make sense.

mala/datahandling/ldos_aligner.py
RandomDefaultUser: For now, I would not worry about MPI, since this is experimental work for the most part, and the mass density project, in which this module is used, is still ongoing. Maybe we add a warning in the beginning that advises users not to trust results if MPI is used with the aligner?

RandomDefaultUser (changes requested):
In general this looks good, thank you @franzpoeschel !

In response to your comments I'd suggest we extend this PR by three things you mentioned:

1. A warning about using MPI
2. A test for numpy vs. openpmd
3. Replacing the numpy direct call with our internal interface

If you don't have the bandwith to work on this at the moment let me know and I can also implement these changes.

PR #654, opened 2025-04-14
Added copy/deepcopy for Network class

### LINKED ISSUE #577: `Network.__new__`'s required `params` argument breaks torch compatibility
When this is a required arg, it breaks `torch.save()` and `torch.load()`. But more importantly, it also breaks code that does `copy.deepcopy(model)`. 

One option is to make it an optional keyword arg which defaults to `None`. When `torch.load()` de-serializes an object, it will first call `__new__` to create a class, then create an instance and populate this with data, such that in this case `params=None` is justified.

RandomDefaultUser:
This PR allows deepcopy on instances of the Network subclasses. It also removes obsolete network classes that are not used anymore, for which more adjustments in the deepcopy would have had to been made. Their deprecation has been long overdue, and while originally planned for v1.4.0, makes sense here as well. Closes #577 .


elcorto:
Thanks, this looks great. 

Have you thought about adding a small test, e.g. create a model, make a copy, run a prediction with both, making sure the results are identical?

RandomDefaultUser:
Good point, I added a test!

mala/network/network.py
```diff
@@ -57,32 +57,27 @@ def __new__(cls, params: Parameters):
         """
         model = None
```
elcorto, line 58: This can be skipped since no test `if model is None` is done in the conditional below.

elcorto (changes requested):
Just a tiny thing, else this looks solid.

mala/network/network.py
RandomDefaultUser: Good point, I deleted this.

RandomDefaultUser:
Thanks for the feedback @elcorto I deleted the line as per your suggestion and will now merge the PR.

PR #656, opened 2025-04-15
Update runner to Ubuntu 22.04

RandomDefaultUser:
As per https://github.com/actions/runner-images/issues/11101.

If I understand correctly, we are already using the 22.04 in two places, and the 20.04 in a different one. I am not sure of the top of my head what the reason for this was, so for now I updated and we'll see what needs fixing.


DanielKotik:
LGTM

DanielKotik (approved):

PR #661, opened 2025-04-29
Renamed "install" to "pipeline"

### LINKED ISSUE #623: "Delete" conda from MALA
We use conda within MALA for the CI. But we do not officially support the conda packages as a way to install MALA. This should be reflected more clearly in the repository structure. Also we should move away from conda to mamba.

RandomDefaultUser:
This renames the folder "install" to "pipeline". The reason for doing this is that the conda files in the "install" folder are not the official way of installing MALA for users. We don't explain the usage in the documentation and don't account for LAMMPS or the TEM. Instead, this folder holds conda files for installing everything for the pipeline. By renaming the folder, we avoid confusion that at this point already some users had. Except for renaming, nothing has changed. 
Closes #623


DanielKotik:
I'm fine with it.

PR #666, opened 2025-05-23
v1.4.0 - Mixology Certification

RandomDefaultUser:
# New features

- Pre- and postprocessing for elemental mixtures added
	- MALA can now natively handle multielement systems (both with bispectrum and ACE descriptors, see below)
	- This feature is still tested in production, so now extensive documentation exists at this point
- Splitting along the energy axis now supported
	- This massively reduces the data cost for systems including core electrons
- ACE descriptors implemented, require specific LAMMPS version to function
	- More details will be described in upcoming paper
	- Currently, now GPU calculation available
- Descriptors can now be calculated on-the-fly 
	- This is especially useful in the GPU-accelerated case, since the GPU calculation of the descriptors is typically very fast and can be done directly before the training
	- Through this functionality, descriptor related hyperparameters can be fixed more easily

# Changes to API/user experience

- DataScaler are now saved in .json format
- MALA tested up to Python 3.13.3

# Fixes

- Fixed bug that broke DDP performance
- Several smaller internal API fixes
- Improved CI


srajama1 (approved):
Thanks @RandomDefaultUser ! I am approving considering the time constraints.

PR #670, opened 2025-05-28
TEM: use ld -z noexecstack

### LINKED ISSUE #667: meson-build TEM fails to import: cannot enable executable stack as shared object requires: Invalid argument (Python 3.13)
After #657 was merged, I tried to build the TEM in a docker container (based on `debian:trixie-20250520`, Python 3.13) using

```
meson     1.8.1
ninja     1.11.1.4
numpy     2.2.6
packaging 25.0
pip       25.1.1
wheel     0.46.1
```

During compilation we see this warning

```
3.019 [5/5] Linking target total_energy.cpython-313-x86_64-linux-gnu.so
3.019 /usr/bin/ld: warning: efermig.o: requires executable stack (because the .note.GNU-stack section is executable
```

and then this seemingly related error when importing

```
3.098 + python3 -c 'import total_energy'
3.111 Traceback (most recent call last):
3.112   File "<string>", line 1, in <module>
3.112     import total_energy
3.112 ImportError: /opt/soft/git/mala/external_modules/total_energy_module/total_energy.cpython-313-x86_64-linux-gnu.so: cannot enable executable stack as shared object requires: Invalid argument
```

A quick search shows that one can solve that using `ld -z noexecstack`, and indeed adding this to `LDFLAGS` removes the warning and the import works.

I have not yet tested this TEM in production, so I can't say whether this fix has side effects.

elcorto:
Closes #667

### DIFF
--- a/external_modules/total_energy_module/build_total_energy_module.sh
+++ b/external_modules/total_energy_module/build_total_energy_module.sh
@@ -71,10 +71,12 @@ cp $root_dir/XClib/xc_lib.a $here/libs/libxclib.a
 # seems.
 ar rcs $tmp_lib_dir/liballobjs.a $pwobjs $utilobjs $modobjs
 
+# LDFLAGS: "-z noexecstack" because of
+# https://github.com/mala-project/mala/issues/667
 FC="$f_compiler" \
 CC="$c_compiler" \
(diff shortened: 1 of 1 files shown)


RandomDefaultUser:
Looks good to me, thanks!

PR #673, opened 2025-06-18
Fix grid dimension type conversion (#669)

### LINKED ISSUE #669: numpy 2: wrong string formatting causes TEM to crash
Running in a container (`debian:bookworm-20250520`, Python 3.11, numpy 2.2.6, TEM built with f2py's meson backend), using the latest `develop` (2ca9b6eea0a98ed72e8c791d87ea6413829ad3c0), there are a number of places with wrong string formatting of numbers when printing them

```
ic| predictor.target_calculator.band_energy: np.float64(106.44888574921316)
```

which is maybe not critical. However the QE input file has this

```
&SYSTEM
   ibrav            = 0
   ecutwfc          = 40.0
   ecutrho          = 160.0
   nr1              = np.int64(36)
   nr2              = np.int64(36)
   nr3              = np.int64(54)
   nr1s             = np.int64(36)
   nr2s             = np.int64(36)
   nr3s             = np.int64(54)
```

which causes QE to fail:

```
Starting QuantumEspresso to get density-based energy contributions.

     Program PWSCF v.7.2 starts on 27May2025 at 16: 1: 4

[...]

     Reading input from mala.pw.scf.f2ahnl86.in

 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
     Error in routine  read_namelists (1):
      bad line in namelist &system: "   nr1              = np.int64(36)" (error could be in the previous line)
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```

elcorto:
Closes #669 

Ran the test suite, no fails that seem to be due to this change. There are some, but due to other bugs that I'll report separately when time permits.

### DIFF
--- a/mala/network/predictor.py
+++ b/mala/network/predictor.py
@@ -114,8 +114,8 @@ class Predictor(Runner):
             # We only need the diagonal elements.
             factor = np.diag(new_cell) / np.diag(old_cell)
             factor = factor.astype(int)
-            self.data.grid_dimension = (
-                factor * self.data.target_calculator.grid_dimensions
+            self.data.grid_dimension = list(
+                map(int, factor * self.data.target_calculator.grid_dimensions)
             )
 
         self.data.grid_size = np.prod(self.data.grid_dimension)


jmgoff (approved):
Quick fix to the bug with the total energy model. Thanks!

PR #674, opened 2025-06-18
Bump mendeleev to 1.1.0 (#672)

### LINKED ISSUE #672: Require mendeleev >= 1.1.0
Since #668 we require mendeleev >= 1.0.0. Version 1.0.0 still [breaks the install with Python 3.11](https://github.com/lmmentel/mendeleev/issues/240). This is [fixed in mendeleev 1.1.0](https://github.com/lmmentel/mendeleev/pull/241). While requiring mendeleev >= 1.0.0 does pull in the newest version, we should require 1.1.0 since that is actually the one which installs correctly with numpy 2 and all Python versions.

elcorto:
Closes #672

### DIFF
--- a/pipeline/mala_cpu_environment.yml
+++ b/pipeline/mala_cpu_environment.yml
@@ -87,7 +87,7 @@ dependencies:
   - markdown=3.8
   - markupsafe=3.0.2
   - matplotlib-base=3.10.3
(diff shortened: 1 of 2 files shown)


RandomDefaultUser:
Looks good to me, thank you!

